PR #46, opened 2020-06-05
Location and room injection for installedapps and devices

bflorian:
Adds a verbose flag for installedapps that includes the location name and for devices that includes the location and room names.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -54,8 +54,10 @@
 		"@oclif/plugin-help": "^2.2.3",
 		"@oclif/plugin-not-found": "^1.2.3",
 		"@smartthings/cli-lib": "^0.0.0",
+		"@types/lodash": "^4.14.155",
 		"aws-sdk": "^2.690.0",
 		"generator-smartthings": "^1.5.0",
+		"lodash": "^4.17.15",
 		"node": "^12.16.2",
 		"open": "^7.0.3",
 		"yeoman-environment": "^2.9.1"
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -54,7 +54,6 @@
 		"@oclif/plugin-help": "^2.2.3",
 		"@oclif/plugin-not-found": "^1.2.3",
 		"@smartthings/cli-lib": "^0.0.0",
-		"@types/lodash": "^4.14.155",
 		"aws-sdk": "^2.690.0",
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/devices.ts
```diff
@@ -51,6 +55,10 @@ export default class DevicesCommand extends ListingOutputAPICommand<Device, Devi
 			description: 'filter results by device',
 			multiple: true,
 		}),
+		verbose: flags.boolean({
+			description: 'include location name in table output',
```
rossiam, line 59: I would drop "in table output" here because this will also show up in json or yaml output.

packages/cli/package.json
```diff
@@ -54,8 +54,10 @@
 		"@oclif/plugin-help": "^2.2.3",
 		"@oclif/plugin-not-found": "^1.2.3",
 		"@smartthings/cli-lib": "^0.0.0",
+		"@types/lodash": "^4.14.155",
```
rossiam, line 57: This should be under "devDependencies".

packages/cli/src/lib/api-helpers.ts
```diff
@@ -0,0 +1,126 @@
...
+
+export async function addLocationsAndRooms(client: SmartThingsClient, list: LocationRoomItem[]): Promise<void> {
+	const locations = await client.locations.list()
+	const locationMap: {[name: string]: string} = locations.reduce((map, obj) => {
+		map[obj.locationId] = obj.name
+		return map
+	}, new IdNameMap())
+
+	// eslint-disable-next-line no-console
+	console.log(`LOCATIONS = ${JSON.stringify(locationMap)}`)
```
rossiam, line 53: Guessing you didn't mean to leave this in but if you did, we should use the logger.

packages/cli/src/lib/api-helpers.ts
```diff
@@ -0,0 +1,126 @@
...
+			if (room.roomId) {
+				map2[room.roomId] = room.name || ''
+			}
+			return map2
+		}, new IdNameMap())
+		return map
+	}, new NestedIdNameMap())
+
+	// eslint-disable-next-line no-console
+	console.log(`LOCATIONS & ROOMS = ${JSON.stringify(locationRoomMap)}`)
```
rossiam, line 79: this too

PR #54, opened 2020-06-16
create rooms command

patshiab:
Created rooms command to display a list of the rooms when called with no id. Users can also call the rooms command with an roomId or index from the table to display a particular room's information. If the command is called with a locationId, the rooms of that location will be listed. 

I also updated the readme for the schema commands, installedapps:delete command, and room command.


packages/cli/src/commands/rooms.ts
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	static args = [{
+		name: 'idOrIndex',
+		description: 'the room id',
+	}]
+
+	primaryKeyName = 'roomId'
+	sortKeyName = 'name'
+
```
rossiam, line 57: We should really be including the locationId in the list output as well. The way you currently do that is to override the `tableHeadings()` method. (You can look at deviceprofiles.ts or capabilities.ts for examples.)

We should also probably support the `--verbose` command line option and include the location name as well when specifying it. Let's write another issue to do that though rather than doing it here so we don't expand the scope of this sprint.

rossiam:
I hadn't thought of this before but this should probably be under locations (as in "locations:rooms" instead of just "rooms") since that's how the API URL is structured. Maybe we could add an alias for just "rooms" (you can see an example of an alias in deviceprofile.ts).

PR #57, opened 2020-06-18
add rooms:update

patshiab:
Added rooms:update command. Users can now update rooms using the roomId, a form of input, and the locationId.

### DIFF
--- a/packages/cli/src/commands/locations/rooms/update.ts
+++ b/packages/cli/src/commands/locations/rooms/update.ts
@@ -1,0 +1,49 @@
+import { flags } from '@oclif/command'
+import { Room, RoomRequest } from '@smartthings/core-sdk'
+
+import { buildTableOutput, getRoomsByLocation } from '../rooms'
+import { SelectingInputOutputAPICommand } from '@smartthings/cli-lib'
+
+
+export default class RoomsUpdateCommand extends SelectingInputOutputAPICommand <RoomRequest, Room, Room> {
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/locations/rooms/update.ts
```diff
@@ -0,0 +1,49 @@
...
+	static flags = {
+		...SelectingInputOutputAPICommand.flags,
+		locationId: flags.string({
+			char: 'l',
+			description: 'a specific locationId to query',
+		}),
+	}
+
+	static args = [{
+		name: 'idOrIndex',
```
rossiam, line 20: This doesn't affect how it works but technically, this is just an `id` for the `Selecting...` classes. (They can't specify an index on the command line--only when prompted.)

packages/cli/src/commands/locations/rooms/update.ts
```diff
@@ -0,0 +1,49 @@
...
+			() => roomsPromise,
+			async (id, data) => {
+				const room = (await roomsPromise).find(room => room.roomId === id)
+				if (!room) {
+					throw Error(`could not find room with id ${id}`)
+				}
+				return this.client.rooms.update(id, data, room.locationId)
+			})
+	}
+}
```
rossiam, line 49: Not a big deal but there is a setting in VSCode that will make VSCode automatically add the new-line at the end of the file for you.

PR #62, opened 2020-06-24
add rules CRUD commands

patshiab:
Added rules CRUD commands:
- Added rules:create command. Users can now create rules using a locationId and a form of input.
- Added rules command to display a list of all rules when called with no rule Id and no locationId. Users can also call the rules command with a rule Id or index from the table to display a particular rule's information. If the command is called with a locationId, the rules of that location will be listed.
- Added rules:update command. Users can now update rules using the rule Id, a form of input, and the locationId.
- Added rules:delete command. Users can now delete rules using the rule Id and locationId. If no rule Id is included, a table containing all of the rules will be outputted and users can choose which rule to be deleted by index or rule Id.

I also updated the arguments for the rooms command to be more accurate.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -127,13 +127,17 @@ Commands are organized in a hierarchy that maps the API hierarchy.
 * [`smartthings locations:delete [ID]`](#smartthings-locationsdelete-id)
 * [`smartthings locations:rooms [IDORINDEX]`](#smartthings-locationsrooms-idorindex)
 * [`smartthings locations:rooms:create`](#smartthings-locationsroomscreate)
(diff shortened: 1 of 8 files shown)


packages/cli/README.md
```diff
@@ -1245,7 +1249,127 @@ ALIASES
 
 _See code: [dist/commands/locations/rooms/update.ts](https://github.com/SmartThingsCommunity/smartthings-cli/blob/v0.0.0/dist/commands/locations/rooms/update.ts)_
 
+<<<<<<< HEAD
```
rossiam, line 1252: Some conflict markers got left in here...I would recommend just regenerating the README with a build.

PR #63, opened 2020-06-26
Fixed device presentation display

bflorian:
Fixes to CLI to address issues with device presentation commands and to align it with future direction of the API

* Corrected _presentation_ command arguments to correctly consist of a required `vid` and options `mnmn`
* Added  _devices:presentation_ command to display the presentation of a device
* Added _deviceprofiles:presentation_ command as a convenience to display the presentation of a profile
* Add -v flag to _deviceprofiles_ command to display `vid` and `mnmn` in the list table

### DIFF
--- a/packages/cli/src/commands/deviceprofiles/presentation.ts
+++ b/packages/cli/src/commands/deviceprofiles/presentation.ts
@@ -1,0 +1,44 @@
+import { DeviceProfile, PresentationDevicePresentation } from '@smartthings/core-sdk'
+
+import { ListingOutputAPICommand } from '@smartthings/cli-lib'
+
+import { buildTableOutput } from '../presentation'
+
+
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/deviceprofiles/presentation.ts
```diff
@@ -0,0 +1,44 @@
...
+
+		this.processNormally(
+			args.id,
+			() => { return this.client.deviceProfiles.list() },
+			async (id) => {
+				const profile = await this.client.deviceProfiles.get(id)
+				if (profile.metadata) {
+					return this.client.presentation.getPresentation(profile.metadata.vid, profile.metadata.mnmn)
+				} else {
+					this.logger.error('No presentation defined for device profile')
```
rossiam, line 37: You should be able to just `throw Error('No presentation...` here and this will log the error and exit for you.

PR #67, opened 2020-07-07
fix dry-run output bug

patshiab:
Fixed the dry-run output bug for the create commands. Users can now see their output in yaml via the yaml flag and save the output to a file via the output flag.

### DIFF
--- a/packages/lib/src/io-command.ts
+++ b/packages/lib/src/io-command.ts
@@ -100,15 +100,15 @@ function formatFromFilename(filename: string): IOFormat {
  *   need to display a list to the user to choose from.
  */
 function writeOutputPrivate<T>(data: T, outputOptions: OutputOptions,
-		buildTableOutput: (data: T) => string, forceTableOutput?: boolean): void {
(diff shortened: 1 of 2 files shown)


packages/lib/src/io-command.ts
```diff
@@ -100,15 +100,17 @@ function formatFromFilename(filename: string): IOFormat {
...
 
-	if (forceTableOutput) {
+	if (forceTableOutput && buildTableOutput) {
 		output = buildTableOutput(data)
 	} else if (outputOptions.format === IOFormat.JSON) {
 		output = JSON.stringify(data, null, outputOptions.indentLevel)
 	} else if (outputOptions.format === IOFormat.YAML) {
 		output = yaml.safeDump(data, { indent: outputOptions.indentLevel })
+	} else if (!buildTableOutput) {
+		throw new Error('did not provide function to build table for output')
```
rossiam, line 113: Let's fall back on JSON output here instead of throwing an error.

Mvz-kvmvl (approved):

PR #68, opened 2020-07-07
update to use controller scope

rossiam:
Updated scopes for PKCE auth flow to new controller scope. Also fixed a couple of minor bugs in implementation.

### DIFF
--- a/packages/lib/src/login-authenticator.ts
+++ b/packages/lib/src/login-authenticator.ts
@@ -26,25 +26,7 @@ export const defaultClientIdProvider: ClientIdProvider = {
 }
 
 // All the scopes the clientId we are using is configured to use.
-const scopes = [
-	'r:devices:*',
-	// 'l:devices',
-	'w:devices:*',
-	'x:devices:*',
-
-	'r:locations:*',
-	'w:locations:*',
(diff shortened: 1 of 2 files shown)


packages/lib/src/login-authenticator.ts
```diff
@@ -205,6 +188,7 @@ export class LoginAuthenticator implements Authenticator {
 		// eslint-disable-next-line no-async-promise-executor
 		return new Promise(async (resolve, reject) => {
 			while (!this.authenticationInfo && Date.now() < startTime + maxDelay) {
+				process.stderr.write('.')
```
john-u, line 191: Is there a functional reason this is `stderr` instead of `stdout`?

rossiam: It's to avoid messing with the real output of the command since we often output JSON or YAML, we don't want to mix messages to the user like this in with those.

This article does a better job than I would explaining it under "4. Mind the streams":

https://medium.com/@jdxcode/12-factor-cli-apps-dd3c227a0e46

john-u: Awesome, thanks for the reference.

PR #69, opened 2020-07-08
add configuration documentation

rossiam:
I added a little more documentation, especially for configuration.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -32,7 +32,22 @@ progress.
 1. Run `smartthings --help` to make sure it's working.
 1. Run a specific command with `smartthings <command>`
 
-# Configuration
+## Input and Output Considerations
+
+Many commands in the CLI handle complex input and/or output, mostly for use with
+the SmartThings REST API. Input can always be passed as JSON or YAML and in
+a couple cases a "question and answer" mode is provided. The output format
+will match the input format unless otherwise specified.
+
+| Name | Shortcut | Description |
+| -- | -- | -- |
+| json | j | Write output in JSON format. |
+| yaml | y | Write output in YAML format. |
+| indent | | Specify the number of spaces for YAML or JSON output |
+| input | i | Specify a filename for input. |
+| output | o | Specify a filename for output. The extension of this file will control its type unless overridden with `--json` or `--yaml`. |
+
+## Configuration
 
 The CLI can be configured by creating a YAML file called `config.yaml` in the
 following location:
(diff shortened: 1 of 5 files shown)


packages/cli/doc/configuration.md
```diff
@@ -0,0 +1,59 @@
+# Configuration
+
+## Overview
+
+The CLI can be configured by creating a YAML file called `config.yaml` in the
+following location:
+
+* `$HOME/.config/@smartthings/cli` on OS/X or Linux
```
john-u, line 8: Maybe we should use `macOS` here since Big Sur 11.0 is on its way to market.

packages/cli/doc/configuration.md
```diff
@@ -0,0 +1,59 @@
...
+
+Logging is configured using the `logging` key in the config file. The value of
+this key is passed directly to log4js so any valid log4js configuration can
+be used here. The following categories are used in the CLI:
+
+* rest-client - This category is used for the SDK that interfaces with the API.
+  Turn this on to see detailed information for HTTP calls are made to SmartThings.
+* cli - This is the generic logger used by the CLI.
+
+There are other categories at more may be added later.
```
john-u, line 59: ~at~ and

packages/cli/README.md
```diff
@@ -76,12 +86,19 @@ features](https://github.com/darvid/trine/wiki/YAML-Primer).
...
-   output data in either format as well.
+1. The CLI accepts data in YAML or JSON format and can
+   output data in either format as well as the default table format.
+1. Commands that take input accept stdin or a file specified
+   by the `--input` (shortcut `-i`) flag.
+1. Commands that output data will output the data to stdout unless a file
+   is specified the using `--output (shortcut `-o`) flag.
+1. When command takes input and results in output, the format of the output
+   will match the input format unless an output filename is specified
+   using `--output` with a different extension.
```
john-u, line 97: Above in "Input and Output Considerations" section we also state 
> The output format will match the input data unless otherwise specified.

It's subtle but might be good to stay consistent.

packages/cli/README.md
```diff
@@ -76,12 +86,19 @@ features](https://github.com/darvid/trine/wiki/YAML-Primer).
...
    	* `smartthings capabilities:presentation:create --help`
-1. In general, the CLI accepts JSON data in YAML or JSON format and will
-   output data in either format as well.
+1. The CLI accepts data in YAML or JSON format and can
+   output data in either format as well as the default table format.
+1. Commands that take input accept stdin or a file specified
+   by the `--input` (shortcut `-i`) flag.
+1. Commands that output data will output the data to stdout unless a file
+   is specified the using `--output (shortcut `-o`) flag.
+1. When command takes input and results in output, the format of the output
```
john-u, line 95: Either this should be 
> When a command. . .
or we make command plural and edit all subject-verb agreement. My only input is we should try to stay consistent and the above list items are plural with commands.

packages/cli/README.md
rossiam: I assume you're talking about me saying "output format" and "input data". No idea what the heck I was thinking...it's like I wrote those two parts of the sentences in different lifetimes. 😄

PR #70, opened 2020-07-08
Updated apps, device profiles, and schema to use latest conventions

bflorian:


packages/cli/src/commands/apps/update.ts
```diff
@@ -32,22 +31,24 @@ export default class AppUpdateCommand extends InputOutputAPICommand<AppRequest,
...
+						if (data.lambdaSmartApp.functions) {
+							const requests = data.lambdaSmartApp.functions.map((it) => {
+								return addPermission(it)
+							})
+							await Promise.all(requests)
+						}
+					} else {
+						this.logger.error('Authorization is not applicable to web-hook SmartApps')
+						// eslint-disable-next-line no-process-exit
+						process.exit(1)
```
rossiam, line 48: You should just be able to `throw Error('Authorization...` instead of logging and exiting and `processNormally` will log the error and exit for you.

packages/cli/src/commands/apps/oauth/generate.ts
```diff
@@ -24,6 +23,8 @@ export default class AppOauthGenerateCommand extends InputOutputAPICommand<AppOA
 		const { args, argv, flags } = this.parse(AppOauthGenerateCommand)
 		await super.setup(args, argv, flags)
 
-		this.processNormally((data) => { return this.client.apps.regenerateOauth(args.id, data) })
+		this.processNormally(args.id,
+			() => { return this.client.apps.list() },
```
rossiam, line 27: When an arrow function has only one command and that command is returned, you can leave off the `return` and the curly brackets. i.e. this line could just be `() => this.client.apps.list(),`

bflorian:
Merging as is and will incorporate suggestions in the next PR

PR #73, opened 2020-07-10
Devices phase1 and selecting output commands

bflorian:
Started bringing device commands up to current practices but ran into issues with multiple, non-ID arguments that I decided to defer to a second PR.

Also switched a number of commands that display non-definition data for objects such as `apps:oauth` from `ListingOutputAPICommand ` to `SelectingOutputAPICommand` for consistency between commands that accept indexes on the command line (because they don't modify anything) and those that don't. Added an argument check to display a descriptive error message when an index is provided for commands that don't support it (because they modify data) rather than just letting the command fail with a 403, which can be confusing.

### DIFF
--- a/packages/cli/src/commands/devices/status.ts
+++ b/packages/cli/src/commands/devices/status.ts
@@ -1,23 +1,31 @@
-import { APICommand } from '@smartthings/cli-lib'
+import { Device, DeviceStatus } from '@smartthings/core-sdk'
+
+import {SelectingOutputAPICommand} from '@smartthings/cli-lib'
 
 
-export default class DevicesStatus extends APICommand {
+export default class DeviceStatusCommand extends SelectingOutputAPICommand<DeviceStatus, Device> {
 	static description = "get the current status of all of a device's component's attributes"
 
-	static flags = APICommand.flags
+	static flags = SelectingOutputAPICommand.flags
 
 	static args = [{
 		name: 'id',
 		description: 'the device id',
-		required: true,
 	}]
 
+	primaryKeyName = 'deviceId'
+	sortKeyName = 'label'
+	listTableFieldDefinitions = ['label', 'name', 'type', 'deviceId']
+	acceptIndexId = true
+
 	async run(): Promise<void> {
-		const { args, argv, flags } = this.parse(DevicesStatus)
+		const { args, argv, flags } = this.parse(DeviceStatusCommand)
 		await super.setup(args, argv, flags)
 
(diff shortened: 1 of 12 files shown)


packages/lib/src/io-command.ts
```diff
@@ -618,14 +649,22 @@ export abstract class SelectingInputOutputAPICommand<I, O, L> extends SelectingI
...
+	protected async processNormally(idOrIndex: ID | string | undefined,
 			listCallback: ListCallback<L>,
 			actionCallback: GetCallback<ID, O>): Promise<void> {
 		try {
 			let inputId: ID
-			if (id) {
-				this.log(`using id from command line = ${id}`)
-				inputId = id
+			if (idOrIndex) {
+				if (this.acceptIndexId || typeof idOrIndex !== 'string' || !idOrIndex.match(validIndex)) {
```
rossiam, line 663: I don't understand the why we're using `typeof` for the middle expression (`|| typeof idOrIndex !== 'string'`). After staring at it for a few minutes, it seems like the purpose of this clause is to pass when no `idOrIndex` is passed but that would be more clear IMO with a simple `|| !idOrIndex`.

bflorian: Without the `typeof idOrIndex !== 'string'` typescript complains about the `idOrIndex.match()` call because `idOrIndex` can either be a string or an object of the `ID` class, and `match()` doesn't accept the latter

packages/cli/src/commands/devices/status.ts
```diff
@@ -1,23 +1,31 @@
-import { APICommand } from '@smartthings/cli-lib'
+import { Device, DeviceStatus } from '@smartthings/core-sdk'
 
+import {SelectingOutputAPICommand} from '@smartthings/cli-lib'
 
-export default class DevicesStatus extends APICommand {
+
+export default class DeviceStatusCommand extends SelectingOutputAPICommand<DeviceStatus, Device> {
```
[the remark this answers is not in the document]
bflorian: Done

packages/lib/src/io-command.ts
bflorian: And it's not a case of no idOrIndex being passed. The outer statement has already tested `if (idOrIndex)`. The idea is to fail when `this.acceptIndexId` is false and `idOrIndex` is an index, so that we get a descriptive error message rather than just the 403 that results from passing an index to the API call

rossiam: Oh, I hadn't thought of the type checking for some reason.

PR #74, opened 2020-07-13
Completed device commands update

bflorian:
Updated to use latest classes and conventions:
* devices:capability-status (renamed to reflect that it accepts only a single capability)
* devices:component-status (renamed to reflect that it accepts only a single component)
* devices:commands

Added:
* devices:rename
* devices:update

Made base class modifications to allow additional arguments to be entered for selecting commands such as capability-status. 

The devices:commands command either accepts input via the `-i` flag or prompts for input in the form:
```
[<component>:]<capability>:<command>[(<arguments>)]
```


packages/cli/src/commands/devices/commands.ts
```diff
@@ -1,49 +1,81 @@
...
+			name: 'deviceCommand',
+			message: 'Enter device command [component:]<capability>:<command>([<arguments]>):',
+			validate: (input: string) => {
+				// switch:on
+				// switch:on()
+				// outlet1:switch:on()
+				// switchLevel:setLevel(20)
+				return new RegExp(/^([a-zA-Z0-9]+:)?[a-zA-Z0-9]+:[a-zA-Z0-9]+(\(.*\))?$/).test(input) || 'Invalid command syntax'
+			},
+		})).deviceCommand
```
rossiam, line 42: I would rather see this asked as multiple questions. If the user has to put it all in one line, why not just have them enter it on the command line?

packages/cli/src/commands/devices/component-status.ts
```diff
@@ -0,0 +1,67 @@
...
+		},
+	]
+
+	protected buildTableOutput = buildTableOutput
+
+	primaryKeyName = 'deviceId'
+	sortKeyName = 'label'
+	listTableFieldDefinitions = ['label', 'name', 'type', 'deviceId']
+	acceptIndexId = true
+	inputPrompt = 'Enter id or index and component'
```
rossiam, line 52: I think this also would benefit from multiple questions.

packages/cli/src/commands/devices/rename.ts
```diff
@@ -0,0 +1,45 @@
...
+		},
+	]
+
+	protected buildTableOutput = buildTableOutput
+
+	primaryKeyName = 'deviceId'
+	sortKeyName = 'label'
+	listTableFieldDefinitions = ['label', 'name', 'type', 'deviceId']
+	acceptIndexId = true
+	inputPrompt = 'Enter <id or index> <new label>'
```
rossiam, line 30: Multiple questions again would be nice.

packages/lib/src/io-command.ts
```diff
@@ -73,10 +74,17 @@ export interface OutputOptions {
 	indentLevel: number
 }
 
+export type InputArgv = string[]
+
+export interface IdAndArgs<ID> {
+	id: ID
+	argv: InputArgv
+}
```
rossiam, line 82: I think it would be easier to read the code if we created a couple of extra specific-purpose "Id" types in the command classes instead (that would then have named properties like "capability" or "component").

bflorian:
Closing this in favor of a rework to address some of the issues and suggestions

PR #77, opened 2020-07-16
chore: replace ci with github actions using lerna

john-u:


john-u:
I just attempted `lerna bootstrap --ci` locally and got a failure due to package.json and package-lock.json out of sync in one of the packages (one of the benefits of using npm ci is to catch this state). I will look into the issue and we can merge this and watch it fail like it should. I can fix the issue in a follow up.

john-u:
Updated to use recommendation[ from lerna FAQ](https://github.com/lerna/lerna/blob/master/FAQ.md#ci-setup) to get us going. Also removed lower level package-locks as lerna manages all packages in the root package-lock when hoisting.

PR #78, opened 2020-07-21
feat(plugins) add support for plugins

rossiam:
* Included @oclif/plugin-plugins plugin so the user can manage plugins vi the "smartthings plugins" command. (I tested this by locally installing the @oclif/plugin-commands plugin.)
* Updated lodash dependency
* fixed a couple bugs with input processing


packages/cli/README.md
```diff
@@ -138,6 +138,11 @@ Commands are organized in a hierarchy that maps to the API hierarchy.
 * [`smartthings locations:rooms:delete [ID]`](#smartthings-locationsroomsdelete-id)
 * [`smartthings locations:rooms:update [ID]`](#smartthings-locationsroomsupdate-id)
 * [`smartthings locations:update [ID]`](#smartthings-locationsupdate-id)
+* [`smartthings plugins`](#smartthings-plugins)
+* [`smartthings plugins:install PLUGIN...`](#smartthings-pluginsinstall-plugin)
+* [`smartthings plugins:link PLUGIN`](#smartthings-pluginslink-plugin)
+* [`smartthings plugins:uninstall PLUGIN...`](#smartthings-pluginsuninstall-plugin)
+* [`smartthings plugins:update`](#smartthings-pluginsupdate)
```
john-u, line 145: These are not part of the SmartThings API hierarchy like the note above suggests.
> Commands are organized in a hierarchy that maps to the API hierarchy.

Should we update that header verbiage a bit to make it clear some commands aren't hitting the API?

packages/cli/README.md
```diff
@@ -1359,6 +1389,128 @@ OPTIONS
   --indent=indent        specify indentation for formatting JSON or YAML output
 ```
 
+<<<<<<< HEAD
+=======
```
john-u, line 1393: Left from a merge conflict?

packages/cli/README.md
```diff
@@ -1359,6 +1389,128 @@ OPTIONS
...
+  $ smartthings plugins:update
+
+OPTIONS
+  -h, --help     show CLI help
+  -v, --verbose
+```
+
+_See code: [@oclif/plugin-plugins](https://github.com/oclif/plugin-plugins/blob/v1.9.0/src/commands/plugins/update.ts)_
+
+>>>>>>> feat(plugins) add support for plugins
```
john-u, line 1513: Seems this is also unintended.

packages/cli/README.md
rossiam: What exactly makes this look like it's part of the API hierarchy? This text here is all auto-generated and the specify lines added here come from an external package so we would have to get them to change the wording or stop auto-generating this if we wanted to change it.

packages/cli/README.md
rossiam: Yeah, I thought I had dealt with that but obviously I hadn't.

rossiam: fixed now

packages/cli/README.md
rossiam: fixed

packages/cli/README.md
john-u: The note above all of these commands explicitly states
> Commands are organized in a hierarchy that maps to the API hierarchy.

Yet, `plugins` and the associated subcommands are an oclif thing and not part of the SmartThings API. There are also other commands in this list that are not part of any SmartThings API. I just thought it might be confusing to some readers. I would not try to edit this section at all unless we could add delimiters or something but maybe state something above like 
> Commands **for SmartThings objects** are organized in a hierarchy that maps to the API hierarchy.

Otherwise what else is telling someone not super familiar with oclif that `smartthings plugins` isn't an API entity?

rossiam: I see what you're saying now. I've updated that text to:

Commands that use the SmartThings REST API are organized in a hierarchy
that maps to that hierarchy.

PR #83, opened 2020-07-27
bug: add missing dependencies

rossiam:
- added missing dependencies
- updated READMEs a little (they still need work)

### DIFF
--- a/packages/lib/README.md
+++ b/packages/lib/README.md
@@ -1,6 +1,10 @@
-# testlib
+# @smartthings/cli-lib
 
-A library to support testing SmartThings CLI plug-ins with Jest.
+A library to support writing SmartThings CLI plug-ins with Jest.
+
+## Creating a plugin
+
+TODO: instructions for setting up a plugin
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
 		"@oclif/command": "^1.5.19",
 		"@oclif/config": "^1.15.1",
 		"@smartthings/core-sdk": "^0.3.0",
+		"cli-table": "^0.3.1",
```
john-u, line 8: Was this done using some form of `lerna add cli-table`? I tried locally, and `lerna` does not seem to add to the root, which leads me to believe it isn't expecting it to work that way and would rather specify in lower packages and hoist them up.

rossiam: I just added it manually and did a `full_clean.sh; bootstrap.sh` I wouldn't expect lerna to add this to the root package.json. Why do you think it should?

rossiam: Weird...I must have put that in accidentally. I've removed it now.

PR #88, opened 2020-08-11
feat: add support for building input form command line options

rossiam:
* added support for getting input from command line options by implementing `getInputFromCommandLine` and `hasCommandLineInput` to `Inputting` mixin
* a little minor clean-up

### DIFF
--- a/packages/lib/src/io-command.ts
+++ b/packages/lib/src/io-command.ts
@@ -54,7 +54,8 @@ const outputFlag = {
 export enum IOFormat {
 	YAML = 'yaml',
 	JSON = 'json',
-	// for input, this is Q & A, for output, it's a human-readable table format
+
+	// for input, this is Q & A or command line, for output, it's a human-readable table format
 	COMMON = 'common',
 }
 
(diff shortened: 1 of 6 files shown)


packages/lib/src/io-command.ts
```diff
@@ -177,16 +205,22 @@ export abstract class Inputting<I> {
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	async setup(args: { [name: string]: any }, argv: string[], flags: { [name: string]: any }): Promise<void> {
-		const hasGetInputFromUser = this.getInputFromUser !== undefined
+		if (this.hasCommandLineInput !== undefined && this.getInputFromCommandLine === undefined
+				|| this.hasCommandLineInput === undefined && this.getInputFromCommandLine !== undefined) {
+			throw new Error('invalid code; implement both hasCommandLineInput and getInputFromCommandLine or neither')
+		}
```
john-u, line 211: It seems like the necessity for this check here and the checking in `readInput` above is a sign we are overloading this Inputting class. Maybe we could refactor into separate strategies since command line input is different than inquirer input etc.

rossiam: If we do that though we will double the number of `...Input.?APICommand` classes that we have. Then, if we also do the same thing with `getInputFromUser`, we would double that number again. I suppose we could clean that up a bit by doing all the mixing-in in the command classes themselves but I'm still a bit leery about letting that out of the bag that way.

john-u: I am not sure why the number of classes would have to double or more depending on the refactor of course. Regardless, I think we've identified a smell here that can be discussed and addressed outside the scope of this PR.

PR #90, opened 2020-08-17
build: update jest, eslint and other dev tools

rossiam:
* updated jest, eslint, TypeScript, etc. packages
* fixed new linting errors as much as I could without introducing more "ignores"


packages/cli/src/commands/deviceprofiles/update.ts
```diff
@@ -49,7 +49,7 @@ export default class DeviceProfileUpdateCommand extends SelectingInputOutputAPIC
 // Cleanup is done so that the result of a device profile get can be modified and
 // used in an update operation without having to delete the status, owner, and
 // component name fields, which aren't accepted in the update API call.
-export function cleanupRequest(deviceProfileRequest: Partial<DeviceProfile>): DeviceProfileRequest {
+export function cleanupRequest(deviceProfileRequest: Partial<DeviceProfile & { restrictions: unknown }>): DeviceProfileRequest {
```
john-u, line 52: Should we create an issue to update the type for this in the SDK?

packages/cli/src/commands/capabilities.ts
```diff
@@ -173,13 +171,12 @@ export async function getStandard(this: APICommand): Promise<CapabilitySummaryWi
 }
 
 export async function getIdFromUser(this: APICommand & { primaryKeyName: string },
-		items: CapabilitySummaryWithNamespace[]): Promise<CapabilityId> {
+		items: (CapabilitySummaryWithNamespace & { [key: string]: unknown })[]): Promise<CapabilityId> {
```
john-u, line 174: Do we need another type?

rossiam: I wouldn't think so for this one of use case but you got me looking at the code again...the only reason this was here was to make `this.primaryKeyName in item` work but since. This function was adapted from a much more generic version in in io-commands but here, `id` can just be hard-coded which makes the code simpler and solves the type error.

I've updated the code to just use `id`. In testing it, I noticed a bug in `capabilities:delete` but I'm going to fix that in another issue because I think the project is far enough along I need to stop doing more than one thing in a single issue, no matter how small. :-)

packages/cli/src/commands/deviceprofiles/update.ts
rossiam: Good question...it's not in the documentation that this field should ever be there but it must be or @bflorian wouldn't have had to clear it here. (I'm guessing it's a computed field.) What do you think Bob?

.eslintrc.js
```diff
@@ -80,6 +80,7 @@ module.exports = {
 		],
 		'@typescript-eslint/no-var-requires': 'error',
 		'@typescript-eslint/ban-ts-ignore': 0,
+		'@typescript-eslint/explicit-module-boundary-types': 'off',
```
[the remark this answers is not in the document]
rossiam: This one is new and weird...it seems to give errors in two cases in our code:

1. Where we are using inline functions where the type can be inferred from context. (We already turned down another rule for this.)
2. When we use `any`. We already have the "no explicit any" rule on so this just forces us to add two ignores to the same line when we do which just seems silly to me.

packages/cli/src/commands/deviceprofiles/update.ts
bflorian: The type def in the core API is correct. It field isn't there. What I wanted to accomplish is to allow the result of a device profile get call to be used in a create call to make a copy of the profile, without having to go through and remove a bunch of fields. So the proper thing would have been to define a new interface local to the CLI that extends the one from the core API and adds those fields as optional

rossiam: It seems like the API _should_ have a separate type for return values since it's not exactly the same as what is accepted. (Then, we would have a separate type in the core SDK as well.)

We _could_ include the extra type even though the API doesn't but I don't love that idea.

.eslintrc.js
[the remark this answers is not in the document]
john-u: For 2, there may be cases where we can start using [unknown](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type). Depending on the number of failures seen, it may be out of scope to fix for this PR.

packages/cli/src/commands/deviceprofiles/update.ts
john-u: > It seems like the API should have a separate type for return values since it's not exactly the same as what is accepted
From what I see it does. We are just trying to avoid having 2 interfaces here for both request and response between endpoints.

There a multiple ways to handle this in clients, one being some sort of adaptation between them as we did here, but this seems convoluted/confusing at first glance. I recommend just creating a ticket with these details here and we can circle back to address.

.eslintrc.js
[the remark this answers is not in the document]
rossiam: The `any` duplication only shows up in one place in our code, this line:

```
export function applyMixins(derivedCtor: any, baseCtors: any[], options?: MixinOptions): void {
```

We can deal with that by either updating the ignore on the preceding line:

```
// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
```

or, by ignoring the `any` part of this rule:

```
		'@typescript-eslint/explicit-module-boundary-types': [
			'error',
			{ allowArgumentsExplicitlyTypedAsAny: true },
		],
```

For that I would just add the extra disable. The thing that really bothers me if we enable this rule is that we have to start adding types to all of our in-line functions where the type can be inferred from the context. Here's a snippet of a diff wit this rule turned on and all the errors fixed:

```
-		include: app => !!app.webhookSmartApp?.publicKey,
-		value: app => app.webhookSmartApp?.publicKey?.replace(/\r\n/g, '\n') ?? '',
+		include: (app: App): boolean => !!app.webhookSmartApp?.publicKey,
+		value: (app: App): string => app.webhookSmartApp?.publicKey?.replace(/\r\n/g, '\n') ?? '',
```

I don't see how this extra verbosity helps anything.

packages/cli/src/commands/deviceprofiles/update.ts
rossiam: We do already have separate objects for the request and response (`DeviceProfileRequest` and `DeviceProfile` respectively)...and they match the API documentation which does not include the `restrictions` field in either one:

https://smartthings.developer.samsung.com/docs/api-ref/st-api.html#operation/createDeviceProfile

(I don't see it in the response for any of the endpoints--not even in the examples.)

.eslintrc.js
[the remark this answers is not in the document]
john-u: I also think it's strange how this specific case can't be disabled for this one.

PR #94, opened 2020-08-20
docs(lib): add more detailed comments on the various APICommand classes

rossiam:
- added more detailed comments on the various APICommand classes

### DIFF
--- a/packages/lib/src/io-command.ts
+++ b/packages/lib/src/io-command.ts
@@ -128,6 +128,15 @@ function writeOutputPrivate<T>(data: T, outputOptions: OutputOptions,
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface Inputting<I> extends Loggable {}
+/**
+ * The "Inputting" mixin is used to add the ability of accepting complex
+ * input in the form of JSON or YAML (from stdin or a file) to a command.
+ *
+ * Additionally, this input can be built via command line if
+ * hasCommandLineInput and getInputFromCommandLine are implemented.
+ * Likewise, implementing getInputFromUser allows a command to also build
+ * input from a Q and A session with the user.
+ */
 export abstract class Inputting<I> {
 	private _inputOptions?: InputOptions
 
@@ -246,8 +255,9 @@ const inputFlags = {
 }
 
 /**
- * Mixin meant to be used along with Outputting and/or Listing that provides
- * instance of OutputOptions.
+ * Since both Outputting and Listing need OutputOptions, this is a simple
+ * mixin that adds it and needs to be included when one or both of those
+ * mixins are used.
  */
 abstract class Outputable {
 	private _outputOptions?: OutputOptions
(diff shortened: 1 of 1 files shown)


packages/lib/src/io-command.ts
```diff
@@ -128,6 +128,15 @@ function writeOutputPrivate<T>(data: T, outputOptions: OutputOptions,
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface Inputting<I> extends Loggable {}
+/**
+ * The "Inputting" mixin is used to add the ability of accepting complex
+ * input in the form of JSON or YAML to a command.
```
john-u, line 133: May want to specify as a filepath/name.

packages/lib/src/io-command.ts
```diff
@@ -246,8 +255,8 @@ const inputFlags = {
 }
 
 /**
- * Mixin meant to be used along with Outputting and/or Listing that provides
- * instance of OutputOptions.
+ * Since both Outputting and Listing need OutputOptions, this is a simple
+ * mixin that needs to be included when one or both of those mixins are used.
```
john-u, line 259: I usually try to use a doc format like `@see` when referencing symbols in comments in case we refactor later. However, I tried with VSCode and couldn't really get it to link to code or refactor with it as well as I have with IntelliJ. So our milage will very with htat.

packages/lib/src/io-command.ts
```diff
@@ -302,6 +311,16 @@ const inputOutputFlags = {
...
 
+/**
+ * The "Outputting" mixin is used to add the ability of presenting complex
+ * output in the form of JSON, YAML or a more human-readable format.
+ * JSON and YAML output are built-in and require no extra implementation but
+ * for the more human-readable output (which is the default), one must
+ * define the tableFieldDefinitions property or override the buildTableOutput
+ * method. See TableFieldDefinition in table-generator.ts for more details.
+ *
+ * NOTE: this mixin requires that Outputable also be mixed in to.
```
john-u, line 322: ~to~ too or just drop to altogether.

packages/lib/src/io-command.ts
```diff
@@ -363,6 +402,11 @@ export abstract class Listing<L> extends Outputable {
 	}
 }
 
+/**
+ * Extend this class for your command if you need to accept input but the
+ * API call you're making doesn't have complex output. There are currently
+ * no examples of this in the CLI (as of August 2020).
+ */
```
john-u, line 409: Should we have it then? Maybe we can clean up in support of [YAGNI](https://martinfowler.com/bliki/Yagni.html).

packages/lib/src/io-command.ts
rossiam: Added the text "(from stdin or a file)".

packages/lib/src/io-command.ts
rossiam: We can look into that more but I'm not sure that tsdoc supports it. I think we should consider something like Sphinx for our docs at some point anyway...I think that will play better with actually getting them posted somewhere and give us a lot more flexibility.

packages/lib/src/io-command.ts
rossiam: fixed

packages/lib/src/io-command.ts
rossiam: If this were in the packages/cli, I would have already removed it but since it's in the packages/lib for plugins to use, I think we should err on the side of being more complete.

PR #97, opened 2020-09-01
feat: export api-helpers functions

rossiam:
* moved api-helpers.ts into `lib` and exported its methods
  * added unit tests
  * modified methods to return new objects rather than update-in-place
  * removed unused room-only method (rooms have to have a location anyway)
* added "test-coverage" script that runs jest with with test coverage reports
* updated JS target to es2019 since the version of node we are targeting supports it and it gives us access to `flat` and `flatMap` array methods


packages/lib/src/__tests__/api-helpers.test.ts
```diff
@@ -0,0 +1,235 @@
+// import { mocked } from 'ts-jest/utils'
```
john-u, line 1: Could remove this, otherwise looks good.

rossiam: fixed

PR #100, opened 2020-09-03
test: add unit tests for logger

rossiam:
* added unit tests for logger.ts
* fixed minor bug with `level` property (get) on a logger that was discovered by tests
* removed a couple of `any` uses

### DIFF
--- a/packages/lib/src/__tests__/logger.test.ts
+++ b/packages/lib/src/__tests__/logger.test.ts
@@ -1,0 +1,136 @@
+import { LoggingEvent } from 'log4js'
+
+import { Logger } from '@smartthings/core-sdk'
+
+import { logManager } from '../logger'
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const recording = require('log4js/lib/appenders/recording')
+
+
+describe('logger', () => {
+	function setupLogManager(level: string): void {
+		const config = {
+			appenders: { main: { type: 'recording' } },
+			categories: { default: { appenders: ['main'], level } },
+		}
+		logManager.init(config)
+	}
+
+	function logOneOfEach(log: Logger): LoggingEvent[] {
+		log.trace('trace message')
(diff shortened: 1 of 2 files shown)


packages/lib/src/__tests__/logger.test.ts
```diff
@@ -0,0 +1,138 @@
...
+			expect(logEvents.length).toBe(3)
+			expect(logEvents.map(event => event.level.levelStr)).toEqual(['WARN', 'ERROR', 'FATAL'])
+		})
+
+		it('log logs correctly', function() {
+			setupLogManager('trace')
+			const log = logManager.getLogger('my-category')
+
+			// We might consider removing this at some point...with log4js it
+			// seems to log at the info level.
```
john-u, line 98: So are we saying that the test for `log` logging correctly in fact logs incorrectly?

john-u (approved):

packages/lib/src/__tests__/logger.test.ts
rossiam: No, what I was really trying to say is that the `log` method that doesn't specify a level was itself a mistake. (What level _should_ it log at?) Logging frameworks commonly have a `log` method where you specify the level as an argument and when I wrote this in the first place, I proxied log4js methods and I believe I thought this was that method, not noticing the lack of a level specified. (Run-on enough for you? :-))

rossiam: I'll update the comment to make this more clear. Also, I would have removed this when I noticed it but it will require removing (or updating it) in the interface in the core SDK which deserves its own issue IMO.

rossiam: You know what, I"m just going to remove the comment entirely and write an issue instead. That's what I should have done in the first place. (Sorry for the reply-barrage. 😄 )

PR #101, opened 2020-09-03
feat: app list options and fixes to app and schema commands

bflorian:
Several small fixes and enhancements:
* Added flags to `apps` list command to accept filter parameters. 
* Fixed bug in `schema:update` where list selection wasn't working.
* Added flags to app and schema AWS lambda authorization to be able to override default principal and statement-id parameters.

### DIFF
--- a/packages/cli/src/commands/apps/create.ts
+++ b/packages/cli/src/commands/apps/create.ts
@@ -6,6 +6,7 @@ import { InputOutputAPICommand } from '@smartthings/cli-lib'
 
 import { tableFieldDefinitions } from '../apps'
 import { addPermission } from '../../lib/aws-utils'
+import { lambdaAuthFlags } from '../../lib/common-flags'
 
 
 export default class AppCreateCommand extends InputOutputAPICommand<AppRequest, AppCreationResponse> {
@@ -15,7 +16,9 @@ export default class AppCreateCommand extends InputOutputAPICommand<AppRequest, 
 		...InputOutputAPICommand.flags,
 		authorize: flags.boolean({
 			description: 'authorize Lambda functions to be called by SmartThings',
-		})}
+		}),
+		...lambdaAuthFlags,
+	}
 
 	protected buildTableOutput(data: AppCreationResponse): string {
 		return this.tableGenerator.buildTableFromItem(data.app, tableFieldDefinitions)
@@ -30,7 +33,7 @@ export default class AppCreateCommand extends InputOutputAPICommand<AppRequest, 
 				if (data.lambdaSmartApp) {
 					if (data.lambdaSmartApp.functions) {
 						const requests = data.lambdaSmartApp.functions.map((it) => {
-							return addPermission(it)
+							return addPermission(it, flags.principal, flags['statement-id'])
 						})
 						await Promise.all(requests)
 					}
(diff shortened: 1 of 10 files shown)


packages/cli/src/lib/aws-utils.ts
```diff
@@ -30,6 +38,10 @@ export async function addPermission(arn: string, principal = '906037444270'): Pr
 	}
 }
 
-export function addSchemaPermission(arn: string): Promise<string> {
-	return addPermission(arn, '148790070172')
+export function addSchemaPermission(arn: string, principal = '148790070172', statementId = 'smartthings'): Promise<string> {
+	if (!principal) {
+		principal = '148790070172'
+	}
```
rossiam, line 44: Here too, `principal` can't be undefined because of the defaulted parameter.

packages/cli/src/lib/aws-utils.ts
```diff
@@ -2,12 +2,20 @@ import AWS from 'aws-sdk'
...
 		return 'Invalid Lambda ARN'
 	}
 
+	if (!principal) {
+		principal = '906037444270'
+	}
+
+	if (!statementId) {
+		statementId = 'smartthings'
+	}
```
rossiam, line 17: You actually don't need this because you have the default values. (Even passing a literal `undefined` as one of the values will use its default.)

packages/cli/src/commands/apps/create.ts
```diff
@@ -15,7 +15,14 @@ export default class AppCreateCommand extends InputOutputAPICommand<AppRequest,
...
 		authorize: flags.boolean({
 			description: 'authorize Lambda functions to be called by SmartThings',
-		})}
+		}),
+		principal: flags.string({
+			description: 'use this principal instead of the default when authorizing lambda functions',
+		}),
+		'statement-id': flags.string({
+			description: 'use this statement id instead of the default when authorizing lambda functions',
+		}),
```
rossiam, line 24: Since these flags are used so often, It might be worth putting these in some common place and `...`ing them in so if we need to update the text, we can just do it in one place.

packages/cli/src/commands/schema/authorize.ts
```diff
@@ -0,0 +1,48 @@
...
+		'$ smartthings apps:authorize arn:aws:lambda:us-east-1:1234567890:function:your-test-app',
+		'',
+		'Note that this command is the same as running the following with the AWS CLI:',
+		'',
+		'$ aws lambda add-permission --region us-east-1 \\',
+		'    --function-name arn:aws:lambda:us-east-1:1234567890:function:your-test-app \\',
+		'    --statement-id smartthings --principal 148790070172 --action lambda:InvokeFunction',
+		'',
+		'It requires your machine to be configured to run the AWS CLI',
+	]
```
rossiam, line 36: I love that you're including these examples. I need to do that more often.

rossiam (approved):

packages/cli/src/lib/aws-utils.ts
bflorian: Ha! I was thinking it would behave like a null, but not setting a flag does return undefined, so I removed these for clarity

packages/cli/src/commands/apps/create.ts
bflorian: Good idea. I moved them

PR #103, opened 2020-09-16
feat: added device profile and capability localizations

bflorian:
Note that these localizations are lists nested in the primary, listable capability and device profile objects and therefore required new command base classes to support selection from lists by index.

### DIFF
--- a/packages/cli/src/commands/capabilities/translations.ts
+++ b/packages/cli/src/commands/capabilities/translations.ts
@@ -1,0 +1,173 @@
+import { flags } from '@oclif/command'
+
+import {CapabilityLocalization, LocaleReference} from '@smartthings/core-sdk'
+
+import { APICommand, NestedListingOutputAPICommandBase, stringTranslateToNestedId } from '@smartthings/cli-lib'
+
+import { capabilityIdOrIndexInputArgs, getCustomByNamespace, translateToId,
+	CapabilityId, CapabilitySummaryWithNamespace } from '../capabilities'
+
+
+export function buildTableOutput(this: APICommand, data: CapabilityLocalization): string {
+	let result = `Tag: ${data.tag}`
+	if (data.attributes) {
+		const table = this.tableGenerator.newOutputTable({head: ['Name','Label','Description', 'Template']})
+		for (const name of Object.keys(data.attributes)) {
+			const attr = data.attributes[name]
+			table.push([name, attr.label, attr.description || '', attr.displayTemplate || ''])
+			if (attr.i18n?.value) {
+				for (const key of Object.keys(attr.i18n.value)) {
+					const entry = attr.i18n.value[key]
+					table.push([`${name}.${key}`, `${entry ? entry['label'] : ''}`, `${entry ? entry['description'] || '' : ''}`, ''])
+				}
+			}
+		}
+		result += '\n\nAttributes:\n' + table.toString()
+	}
+	if (data.commands) {
+		const table = this.tableGenerator.newOutputTable({head: ['Name','Label','Description']})
+		for (const name of Object.keys(data.commands)) {
+			const cmd = data.commands[name]
+			table.push([name, cmd.label || '', cmd.description || ''])
+			if (cmd.arguments) {
+				for (const key of Object.keys(cmd.arguments)) {
+					const entry = cmd.arguments[key]
+					table.push([`${name}.${key}`, `${entry ? entry['label'] : ''}`, `${entry ? entry['description'] || '' : ''}`])
+				}
+			}
+		}
+		result += '\n\nCommands:\n' + table.toString()
+	}
+	return result
+}
+
+export type CapabilitySummaryWithLocales = CapabilitySummaryWithNamespace & { locales?: string }
+
+export default class CapabilityTranslationsCommand extends NestedListingOutputAPICommandBase<CapabilityId, string, CapabilityLocalization, CapabilitySummaryWithLocales, LocaleReference> {
+
+	static description = 'Get list of locales supported by the capability'
+
+	static flags = {
+		...NestedListingOutputAPICommandBase.flags,
+		namespace: flags.string({
+			char: 'n',
+			description: 'a specific namespace to query; will use all by default',
+		}),
+		verbose: flags.boolean({
+			description: 'include list of supported locales in table output',
+			char: 'v',
+		}),
+	}
+
+	static args = [...capabilityIdOrIndexInputArgs, {
+		name: 'tag',
+		description: 'the locale tag',
+	}]
+
+	static examples = [
+		'$ smartthings capabilities:translations',
+		'┌───┬─────────────────────────────┬─────────┬──────────┐',
+		'│ # │ Id                          │ Version │ Status   │',
+		'├───┼─────────────────────────────┼─────────┼──────────┤',
+		'│ 1 │ custom1.outputModulation    │ 1       │ proposed │',
+		'│ 2 │ custom1.outputVoltage       │ 1       │ proposed │',
+		'└───┴─────────────────────────────┴─────────┴──────────┘',
+		'outputModulation (master)$ st capabilities:translations -v',
+		'┌───┬─────────────────────────────┬─────────┬──────────┬────────────┐',
+		'│ # │ Id                          │ Version │ Status   │ Locales    │',
+		'├───┼─────────────────────────────┼─────────┼──────────┼────────────┤',
+		'│ 1 │ custom1.outputModulation    │ 1       │ proposed │ ko, en, es │',
+		'│ 2 │ custom1.outputVoltage       │ 1       │ proposed │ en         │',
+		'└───┴─────────────────────────────┴─────────┴──────────┴────────────┘',
+		'',
+		'outputModulation (master)$ st capabilities:translations 1',
+		'outputModulation (master)$ st capabilities:translations custom1.outputModulation',
+		'┌───┬─────┐',
+		'│ # │ Tag │',
+		'├───┼─────┤',
+		'│ 1 │ en  │',
+		'│ 2 │ ko  │',
+		'└───┴─────┘',
+		'',
+		'$ smartthings capabilities:translations 1 1',
+		'$ smartthings capabilities:translations 1 en',
+		'$ smartthings capabilities:translations custom1.outputModulation 1 1',
+		'$ smartthings capabilities:translations custom1.outputModulation 1 en',
+		'$ smartthings capabilities:translations custom1.outputModulation en',
+		'Tag: en',
+		'',
+		'Attributes:',
+		'┌────────────────────────┬───────────────────┬────────────────────────────────┬────────────────────────────────────────────────────┐',
+		'│ Name                   │ Label             │ Description                    │ Template                                           │',
+		'├────────────────────────┼───────────────────┼────────────────────────────────┼────────────────────────────────────────────────────┤',
+		'│ outputModulation       │ Output Modulation │ Power supply output modulation │ The {{attribute}} of {{device.label}} is {{value}} │',
+		'│ outputModulation.50hz  │ 50 Hz             │                                │                                                    │',
+		'│ outputModulation.60hz  │ 60 Hz             │                                │                                                    │',
+		'└────────────────────────┴───────────────────┴────────────────────────────────┴────────────────────────────────────────────────────┘',
+		'',
+		'Commands:',
+		'┌──────────────────────────────────────┬───────────────────────┬──────────────────────────────────────────────────┐',
+		'│ Name                                 │ Label                 │ Description                                      │',
+		'├──────────────────────────────────────┼───────────────────────┼──────────────────────────────────────────────────┤',
+		'│ setOutputModulation                  │ Set Output Modulation │ Set the output modulation to the specified value │',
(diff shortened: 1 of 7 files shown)


packages/lib/src/io-command.ts
```diff
@@ -653,6 +659,27 @@ async function stringTranslateToId<L>(this: APICommand & { readonly primaryKeyNa
...
+	let pk2 = idOrIndex2
+	if (isIndexArgument(idOrIndex2)) {
+		const index = Number.parseInt(idOrIndex2)
+		const items = this.sort2(await listFunction(id))
+		const matchingItem: L2 = items[index - 1]
+		if (!(this.primaryKeyName2 in matchingItem)) {
+			throw Error(`did not find key ${this.primaryKeyName2} in data ${JSON.stringify(items)}`)
+		}
+
+		// @ts-ignore
```
john-u, line 676: If we cannot find ways around these, I suggest we start commenting which error(s) we are suppressing [as recommended by Microsoft](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#suppress-errors-in-ts-files-using--ts-ignore-comments). This will make it easier to debug later if something further unexpected starts happening on these lines.

This comment applies for all 7 usages of `@ts-ignore` in this diff.

packages/cli/src/commands/deviceprofiles/translations.ts
```diff
@@ -0,0 +1,134 @@
...
+			table.push([name, component.label, component.description || ''])
+		}
+		result += '\n' + table.toString()
+	}
+	return result
+}
+
+export type DeviceProfileWithLocales = DeviceProfile & { locales?: string }
+
+export default class CapabilityTranslationsCommand extends NestedListingOutputAPICommand2<DeviceProfileTranslations, DeviceProfileWithLocales, LocaleReference> {
```
john-u, line 23: This should be ~Capability~ DeviceProfileTranslationsCommand.

packages/cli/src/commands/deviceprofiles/translations/delete.ts
```diff
@@ -0,0 +1,68 @@
+import {DeviceProfile, LocaleReference} from '@smartthings/core-sdk'
+
+import { NestedSelectingAPICommand2 } from '@smartthings/cli-lib'
+
+
+export default class DeviceProfileTranslationDeleteCommand extends NestedSelectingAPICommand2<DeviceProfile, LocaleReference> {
```
john-u, line 6: `DeviceProfileTranslation` (using singular) is inconsistent from other command names added in this change.

packages/cli/src/commands/deviceprofiles/translations/delete.ts
```diff
@@ -0,0 +1,68 @@
...
+	sortKeyName = 'name'
+	primaryKeyName2 = 'tag'
+	sortKeyName2 = 'tag'
+	listTableFieldDefinitions = ['name', 'status', 'id']
+	listTableFieldDefinitions2 = ['tag']
+
+	// static examples = [
+	// 	'$ smartthings deviceprofiles:delete 63b8c91e-9686-4c43-9afb-fbd9f77e3bb0  # delete profile with this UUID',
+	// 	'$ smartthings deviceprofiles:delete 5                                     # delete the 5th profile in the list',
+	// ]
```
john-u, line 54: We should remove this.

packages/cli/src/commands/deviceprofiles/translations/upsert.ts
```diff
@@ -0,0 +1,74 @@
+import {DeviceProfile, DeviceProfileTranslations} from '@smartthings/core-sdk'
+import {SelectingInputOutputAPICommand} from '@smartthings/cli-lib'
+
+import {buildTableOutput} from '../translations'
+
+
+export default class CapabilityTranslationsUpsertCommand extends SelectingInputOutputAPICommand<DeviceProfileTranslations, DeviceProfileTranslations, DeviceProfile> {
```
john-u, line 7: This should be ~Capability~ DeviceProfileTranslationsUpsertCommand.

packages/lib/src/io-command.ts
```diff
@@ -939,5 +966,298 @@ export abstract class SelectingInputAPICommand<I, L> extends SelectingInputAPICo
...
+/**
+ * Use this as your base when you need to be able to list or get a single
+ * resource from a list originating from another resource, which itself can be listed
+ *
+ * This class accepts the identifier types of both the primary and secondary lists as the ID and ID2 generic inputs.
+ * In most cases, your ids will be a simple strings and you can use NestedListingOutputAPICommand2 class,
+ * which is a little simpler. If the first ID is not a string but the second ID is, you can use the
+ * NestedListingOutputAPICommand class.
+ */
+export abstract class NestedListingOutputAPICommandBase<ID, ID2, O, L, L2> extends APICommand {
```
rossiam, line 980: Some of the duplication could be reduced by making a "Nested" mixin.

rossiam, line 980: I know it's longer but would use `NID` and `NL` for these types and names like `nestedPrimaryKeyName` and `nestedIdOrIndex` instead of just the number "2".

packages/lib/src/io-command.ts
```diff
@@ -939,5 +966,298 @@ export abstract class SelectingInputAPICommand<I, L> extends SelectingInputAPICo
...
+
+/**
+ * Use this as your base when you need to be able to list or get a single
+ * resource from a list that is itself a property of another listable resource.
+ *
+ * This class is used for commands where both identifiers are simple strings.
+ * If you need a more complex identifier type, you can use either the
+ * NestedListingOutputAPICommand or NestedListingOutputAPICommandBase class.
+ */
+export abstract class NestedListingOutputAPICommand2<O, L, L2> extends NestedListingOutputAPICommand<string, O, L, L2> {
```
rossiam, line 1073: I'm trying to figure out a good way to name these without using numbers. Some questions to ask:

1. Which of these two classes will be more common, if any.
    * If one is definitely going to be the most used one, that one should get the name `NestedListingOutputAPICommand`.
   * If neither is the "normal" one, the one that uses both strings probably should get the simplest name (`NestedListingOutputAPICommand`) because that seems more consistent with the other classes we've done to me.
2. If one is less common, should we consider just dropping it altogether specifying `string` and exporting `stringTranslateToId` so commands can add the `protected translateToId = stringTranslateToId` themselves. (I should have exported `stringTranslateToId` anyway.)

bflorian: It's 50:50 right now -- capabilities and device profiles. Capabilities are the only things that require the use of these base classes, because of their id/version identifiers, so I guess both IDs being strings might be more common, but I don't know if we will run into this sort of structure again or not. I elected to use the "2" suffix for the classes where both IDs were strings. Hypothetically there could also be a variant where the first ID is a string and the second is not, though it seems unlikely. I'll look into your suggestion 2

packages/cli/src/commands/capabilities/translations.ts
```diff
@@ -0,0 +1,170 @@
...
+		if (args.id) {
+			if (isNaN(args.id)) {
+				// capability ID is not an index, next param must be the version
+				capabilityIdOrIndex = {id: args.id, version: args.version}
+				tagOrIndex = args.tag
+			} else {
+				// capability ID is an index, then version param, if set, must hold the local
+				tagOrIndex = args.version
+			}
+		}
```
rossiam, line 147: If we want to be consistent with other capabilities commands we need to change this a bit. In this code from "capabilities.ts":
```js
		const idOrIndex = args.version
			? { id: args.id, version: args.version }
			: args.id
```
When there is no version specified, that doesn't necessarily mean `args.id` is an index...it could still be an id just without specifying the version. The `translateToId` function just fills in `1` for now but the intent is to use the most recent version when versions are available. So, using `isNaN` makes this method behave differently.

Since the `tag` can't be numeric and `version` can, maybe we can use a combination of the number of arguments and the "numberness" of the _last_ one (if there are two) to determine this. Some psuedo code for what I'm thinking:

* If there are 3 arguments, they are id, version, tag.
* If there is 1 argument, it is assigned to `capabilityIdOrIndex` (and `translateToId` will handle whether it's an index or an id that has to have a default version added).
* If there are two arguments:
   * If the second one is a number, it's `id`, `version`
   * otherwise, it's `id`, `tag`

(Also "locale" is missing an "e" in the second comment.)

packages/lib/src/io-command.ts
bflorian: Fixed

packages/cli/src/commands/deviceprofiles/translations/delete.ts
bflorian: Fixed

packages/cli/src/commands/deviceprofiles/translations/upsert.ts
bflorian: Fixed

packages/lib/src/io-command.ts
```diff
@@ -653,6 +659,27 @@ async function stringTranslateToId<L>(this: APICommand & { readonly primaryKeyNa
...
+	if (isIndexArgument(idOrIndex2)) {
+		const index = Number.parseInt(idOrIndex2)
+		const items = this.sort2(await listFunction(id))
+		const matchingItem: L2 = items[index - 1]
+		if (!(this.primaryKeyName2 in matchingItem)) {
+			throw Error(`did not find key ${this.primaryKeyName2} in data ${JSON.stringify(items)}`)
+		}
+
+		// @ts-ignore
+		pk2 = matchingItem[this.primaryKeyName2]
```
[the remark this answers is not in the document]
bflorian: Fixed

packages/lib/src/io-command.ts
```diff
@@ -939,5 +966,298 @@ export abstract class SelectingInputAPICommand<I, L> extends SelectingInputAPICo
...
+}
+
+async function stringGetIdFromUser2<L2>(this: APICommand & { readonly primaryKeyName2: string }, items: L2[]): Promise<string> {
+	const convertToId = (itemIdOrIndex: string): string | false => {
+		if (itemIdOrIndex.length === 0) {
+			return false
+		}
+		const matchingItem = items.find((item) => {
+			// @ts-ignore
+			return (this.primaryKeyName2 in item) && itemIdOrIndex === item[this.primaryKeyName2]
```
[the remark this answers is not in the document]
bflorian: Fixed

packages/lib/src/io-command.ts
```diff
@@ -939,5 +966,298 @@ export abstract class SelectingInputAPICommand<I, L> extends SelectingInputAPICo
...
+	if (inputId === false) {
+		throw Error(`unable to convert ${itemIdOrIndex} to id`)
+	}
+	return inputId
+}
+
+/**
+ * This class is used for command like "delete" that act on a resource from a list
+ * property on another listable resource. See SelectingAPICommandBase for more
+ * description of the behacvior of the class
```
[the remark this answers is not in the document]
bflorian: Fixed

packages/lib/src/io-command.ts
```diff
@@ -939,5 +966,298 @@ export abstract class SelectingInputAPICommand<I, L> extends SelectingInputAPICo
...
+ * Use this as your base when you need to be able to list or get a single
+ * resource from a list that is itself a property of another listable resource.
+ *
+ * This class is used for commands where both identifiers are simple strings.
+ * If you need a more complex identifier type, you can use either the
+ * NestedListingOutputAPICommand or NestedListingOutputAPICommandBase class.
+ */
+export abstract class NestedListingOutputAPICommand2<O, L, L2> extends NestedListingOutputAPICommand<string, O, L, L2> {
+	protected translateToId = stringTranslateToId
+	static flags = NestedListingOutputAPICommandBase.flags
```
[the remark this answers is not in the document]
bflorian: Fixed

packages/cli/src/commands/capabilities/translations.ts
bflorian: Hmmm, I did not realize that we were defaulting the version to 1 when the capability ID is specified as a string. If we are going to do that then I think we need to output the version in the default table view of the capability. As it is now you don't see which version is being shown:
```
smartthings capabilities xyz.outputVoltage

Capability: xyz.outputVoltage

Attributes: ...
```

This also means that we have an ambiguity, since the nested tag can also be specified as an index, so 
```
$ smartthings capabilities:translations xyz.outputVoltage 2
```
could either mean version 2 of the capability or the second tag of the default version of the capability. I suppose we could drop indexed support for the second parameter, since in this case at least the tags are very short. But other applications might not be. Or we could leave it but require the version in order to reference the tag by index, i.e.
```
$ smartthings capabilities:translations 3 2                    #second  tag of the third capability/version
$ smartthings capabilities:translations xyz.outputVoltage 2    #version two of the capability
$ smartthings capabilities:translations xyz.outputVoltage es   #Spanish tag of the default capability version
$ smartthings capabilities:translations xyz.outputVoltage 2 3  #third tag of version two of the capability
$ smartthings capabilities:translations xyz.outputVoltage 2 es #Spanish tag of version two of the capability
```

bflorian: Another consideration is that defaulting the capability version when not specified gives us no way to list all the versions of a capability from the CLI, once we do have multiple versions

rossiam (approved):

PR #106, opened 2020-09-18
feat: use separate file for logging config

rossiam:
* use separate logging.xml file for logging configuration instead of reading key from config.yaml file
* removed plain dependencies from main package.json (Lerna doesn't use these anyway) so we don't have to maintain them
* fixed dependency versions for core SDK (they all match now)

### DIFF
--- a/packages/lib/src/logger.ts
+++ b/packages/lib/src/logger.ts
@@ -1,4 +1,6 @@
-import { configure, Configuration, Logger, Level } from 'log4js'
+import fs from 'fs'
+import yaml from 'js-yaml'
+import { configure, Configuration as LoggingConfig, Logger, Level } from 'log4js'
 
 import { Logger as APILogger } from '@smartthings/core-sdk'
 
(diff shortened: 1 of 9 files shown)


packages/lib/src/logger.ts
```diff
@@ -86,6 +88,34 @@ export class LogManager {
...
+
+
+const loggingDocsLink = 'https://github.com/SmartThingsCommunity/' +
+	'smartthings-cli/blob/master/packages/cli/doc/configuration.md#logging'
+export function loadLoggingConfig(filename: string): LoggingConfig {
+	if (!fs.existsSync(filename)) {
+		return defaultLoggingConfig
+	}
+
+	const parsed = yaml.safeLoad(fs.readFileSync(`${filename}`, 'utf-8'))
```
bflorian, line 112: Can't the first argument just be `filename`? Does it have to be a template literal here for some reason?

rossiam: It shouldn't be...probably copied code from where one was and didn't notice I removed everything except the `filename`. Fixed.

PR #120, opened 2020-10-23
test: add unit tests for output-builder.ts

rossiam:
* added unit tests for output-builder.ts functionality
* fixed bug where indent was not checked from config file

### DIFF
--- a/packages/lib/src/__tests__/input-builder.test.ts
+++ b/packages/lib/src/__tests__/input-builder.test.ts
@@ -5,19 +5,23 @@ import { buildInputProcessor } from '../input-builder'
 import { IOFormat } from '../io-util'
 import { SmartThingsCommandInterface } from '../smartthings-command'
 import { DefaultTableGenerator } from '../table-generator'
+
 import { SimpleType } from './io-util.test'
+
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function buildMockCommand(flags: { [name: string]: any } = {}, profileConfig: { [name: string]: any } = {}): SmartThingsCommandInterface {
+	return {
+		logger: new NoLogLogger(),
+		flags,
+		profileConfig,
+		tableGenerator: new DefaultTableGenerator(true),
+	}
+}
 
 
 jest.mock('../input')
 jest.mock('../smartthings-command')
-
-function buildMockCommand(): SmartThingsCommandInterface {
-	return {
-		logger: new NoLogLogger(),
-		flags: {},
-		tableGenerator: new DefaultTableGenerator(true),
-	}
-}
 
 afterEach(() => {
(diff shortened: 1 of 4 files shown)


packages/lib/src/__tests__/output-builder.test.ts
```diff
@@ -1,22 +1,138 @@
...
+	it('gets format from filename with input file when not specified', () => {
+		formatFromFilenameSpy.mockReturnValue(ioUtil.IOFormat.JSON)
+		command.flags.output = 'fn.json'
+		buildOutputFormatter<SimpleType>(command)
+		expect(formatFromFilenameSpy).toHaveBeenCalledTimes(1)
+		expect(formatFromFilenameSpy).toHaveBeenCalledWith('fn.json')
+		expect(jsonFormatterSpy).toHaveBeenCalledTimes(1)
+		expect(yamlFormatterSpy).toHaveBeenCalledTimes(0)
+
+		jest.resetAllMocks()
```
john-u, line 61: It would be valuable in some of these cases to split into separate tests. For example, a JSON case and YAML case. It provides more context with failing tests (ex. both are failing or one is) and narrows time to find issues.

packages/lib/src/__tests__/input-builder.test.ts
```diff
@@ -5,20 +5,24 @@ import { buildInputProcessor } from '../input-builder'
...
-import { SimpleType } from './io-util.test'
 
+import { SimpleType } from './io-util.test'
 
-jest.mock('../input')
-jest.mock('../smartthings-command')
 
-function buildMockCommand(): SmartThingsCommandInterface {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function buildMockCommand(flags: { [name: string]: any } = {}, profileConfig: { [name: string]: any } = {}): SmartThingsCommandInterface {
```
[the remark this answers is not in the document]
rossiam: Maybe. I wasn't sure yet though which is why I didn't put it there yet.

The main thing that gave me pause with regards to putting this in `testlib` was that then we would then need to make `lib` (this code) depend on `testlib` which would create a circular dependency between the two. I don't know if that's a bad thing but it definitely makes me uneasy.

I see `testlib` as more of a library for writing tests for commands (i.e. the `cli` package) rather than this code (i.e. the `lib` package), though there will be some overlap.

packages/lib/src/__tests__/output-builder.test.ts
rossiam: I actually just deleted the second one instead...it was a duplicate rather than a separate test. The tests for the different file types are already taken care of (correctly) by unit tests for `formatFromFilename` itself.

john-u (approved):

PR #123, opened 2020-11-03
feat(logger): default log file path to oclif cacheDir

john-u:
* write `smartthings.log` to [oclif CLI cache directory](https://oclif.io/docs/config) by default
* enforce max file size of 1MB to default log
* resolve a couple of lerna/jest related warnings

### DIFF
--- a/packages/lib/src/logger.ts
+++ b/packages/lib/src/logger.ts
@@ -5,6 +5,10 @@ import { configure, Configuration as LoggingConfig, Logger, Level } from 'log4js
 import { Logger as APILogger } from '@smartthings/core-sdk'
 
 
+const DEFAULT_LOG_FILE_SIZE = 1_000_000 // bytes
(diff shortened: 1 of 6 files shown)


packages/lib/src/logger.ts
```diff
@@ -5,6 +5,10 @@ import { configure, Configuration as LoggingConfig, Logger, Level } from 'log4js
 import { Logger as APILogger } from '@smartthings/core-sdk'
 
 
+const DEFAULT_LOG_FILE_SIZE = 1048576 // bytes
```
rossiam, line 8: Super trivial but you could make this is little more readable by using something like `1024 * 1024`, or, since we probably don't need a full MiB, we could use a MB and use underscores since it's so long: `1_000_000`.

rossiam (approved):

packages/lib/src/logger.ts
john-u: I like the numeric separator.

PR #124, opened 2020-12-01
fix: validate capability command and attribute names

rossiam:
* Added validation for both attribute and command names, requiring them to be max 36 characters, only letters and start with a lowercase letter.
* If the user enters an attribute name longer than 33 characters, allow them to enter a shorter one or acknowledge they will not need a setter since prepending the attribute name with `set` to make a setter name would violate the requirements for the name of the command for the setter.

### DIFF
--- a/packages/cli/src/commands/capabilities/create.ts
+++ b/packages/cli/src/commands/capabilities/create.ts
@@ -23,6 +23,12 @@ const enum Type {
 	BOOLEAN = 'boolean',
 }
 
+const attributeAndCommandNamePattern = /^[a-z][a-zA-Z]{0,35}$/
+function commandOrAttributeNameValidator(input: string): boolean | string {
+	return !!attributeAndCommandNamePattern.exec(input)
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/capabilities/create.ts
```diff
@@ -123,6 +129,10 @@ export default class CapabilitiesCreate extends InputOutputAPICommand<Capability
 				type: 'input',
 				name: 'basicCommandName',
 				message,
+				validate: (input) => {
+					// empty string is allowed here because it ends basic command name input
+					return !input || commandOrAttributeNameValidator(input)
```
john-u, line 134: Does the input check and the corresponding comment make more sense in the validator function?

rossiam: This is the only place were input is optional; all other uses of `commandOrAttributeNameValidator` require input.

john-u (approved):

PR #128, opened 2020-12-10
fix: use InstalledApp fields in installedapps:rename and refactor to use new functional paradigm

rossiam:
* fixed installedapps:rename command (was trying to use output copied from devices)
* added support for "selecting" commands (with and without complex output)
* updated the following commands to use new functional style:
    * installedapps
    * installedapps:delete
    * installedapps:rename
* fixed listing-io to use functions rather than duplicating their functionality
* made `listing*` commands sort properly (including previously-converted `namespaces` command which now needs to specify the field to sort on


packages/cli/src/commands/installedapps/delete.ts
```diff
@@ -1,12 +1,12 @@
...
-import { SelectingAPICommand } from '@smartthings/cli-lib'
+import { selectAndActOn, APICommand } from '@smartthings/cli-lib'
 
 
-export default class InstalledAppDeleteCommand extends SelectingAPICommand<InstalledApp> {
+export default class InstalledAppDeleteCommand extends APICommand {
 	static description = 'delete the installed app instance'
 
-	static flags = SelectingAPICommand.flags
+	static flags = APICommand.flags
```
rossiam, line 9: `selectAndActOn` doesn't contribute any flags. I wasn't sure if I should include an empty map that could be added here for the sake of consistency or not. My biggest worry with leaving it out would be someone cutting and pasting a command that doesn't use it and then updating it to one that does and not remembering to the flags. What do you guys think?

packages/lib/src/acting-io.ts
```diff
@@ -0,0 +1,170 @@
...
+import { formatAndWriteItem, outputList, CommonListOutputProducer, CommonOutputProducer, ActionFunction, ListDataFunction,
+	Sorting, IdRetrievalFunction } from './basic-io'
+import { SmartThingsCommandInterface } from './smartthings-command'
+
+
+// Functions in this file support commands that act on an item. The biggest difference
+// between these and the methods from listing-io.ts is that they list items and immediately
+// query the user for an item to act on (if one wasn't specified on the command line). This
+// makes them safe to use for actions that make changes to the item or delete it.
+
```
rossiam, line 13: Commands in this file don't yet support the `acceptIndexId` implemented in the old version. I figured I would wait to implement that until I we translate a command that uses it which will give me a better feel of what a good API for it will be.

packages/lib/src/acting-io.ts
```diff
@@ -0,0 +1,170 @@
...
+	} else {
+		const list = await outputList(command, listFunction, true)
+		if (list.length === 0) {
+			// Nothing was found; user was already notified.
+			command.exit(0)
+		}
+		computedId = await getIdFromUser(command.primaryKeyName, list)
+	}
+	const updatedItem = await actionFunction(computedId)
+	return [computedId, updatedItem]
```
rossiam, line 99: I'm generally not a fan of tuples (with their unnamed fields) but they seem kind of perfect for this use-case. Anyone else have opinions on this?

packages/lib/src/basic-io.ts
```diff
@@ -12,23 +12,35 @@ import { TableFieldDefinition } from './table-generator'
...
 
 
+export interface Sorting {
+	primaryKeyName: string
+	sortKeyName: string
+}
+export interface Naming {
+	itemName?: string
+	pluralItemName?: string
+}
```
rossiam, line 27: The "nothing found" messages before were just "no items" found which is all fine and dandy if you're just querying, say `smartthings devices` or `smartthings installedapps` and then get "no items found" but I was worried it would be too confusing when you `smartthings installedapps:rename` (without specifying an item to rename) and then get `no items found`.

packages/lib/src/acting-io.ts
john-u: It works and is type safe thanks to TS. I say 👍 .

packages/lib/src/acting-io.ts
```diff
@@ -0,0 +1,170 @@
...
+	return computedId
+}
+
+
+export type SelectingOutputCommand<O, L> = SelectingCommand<L> & CommonOutputProducer<O>
+
+/**
+ * Process a command that selects an item (e.g. a device), performs some action on that device
+ * and then outputs the results. Unless your ids are not simple strings, you probably want to
+ * use `selectActOnAndOutput` below instead.
```
john-u, line 133: Do we have any use cases where ids aren't simple strings? Maybe we could just not export this.

packages/lib/src/basic-io.ts
john-u: Since it matches the behavior of other commands, I think it is fine to say no items found when renaming. The user can always reference the `--help` output. If it helps remove some complexity from functions having to deal with this special type.

packages/lib/src/acting-io.ts
rossiam: Good idea...done.

packages/lib/src/basic-io.ts
rossiam: Are you saying you think I should remove this code that allows us a more specific message or did you misunderstand the reason behind my comment?

packages/cli/src/commands/installedapps/rename.ts
```diff
@@ -1,14 +1,18 @@
...
 
 import { InstalledApp } from '@smartthings/core-sdk'
-import { SelectingOutputAPICommand } from '@smartthings/cli-lib'
-import { buildTableOutput } from '../devices'
+import { selectActOnAndOutput, APICommand } from '@smartthings/cli-lib'
+import { listTableFieldDefinitions, tableFieldDefinitions } from '../installedapps'
 
 
-export default class DeviceComponentStatusCommand extends SelectingOutputAPICommand<InstalledApp, InstalledApp> {
+// TODO: test none found!!!
```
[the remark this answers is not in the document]
rossiam: Oops...that was just to be a reminder to myself to check the case where you don't specify an installed app and there are none...should have removed that before committing.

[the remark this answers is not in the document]
rossiam: (Tested and removed now.)

packages/lib/src/acting-io.ts
rossiam: Capabilities have a two part id with a string and a version. It's possible that might change and then we can remove this but for now we do.

packages/lib/src/basic-io.ts
john-u: I think we should remove the extra naming code. I don't see a strong case for it when we already have several cases using `no items found`, so it feels consistent to me.

packages/cli/src/commands/installedapps/delete.ts
john-u: I think we are already relying on devs to remember to include flags for any new command even if not copying pasting, so this isn't some new risk. Adding an empty map is fine but would start cluttering the otherwise simple function, so I vote to leave it. We should clearly state somewhere that anyone building a new command will need to ensure flags are set accordingly.

john-u (approved):

packages/lib/src/basic-io.ts
rossiam: I wrote it in such a way that we can start including `itemName` (and optionally `pluralItemName` so we get more specific messages in other places very easily as well. Here's what this looks like in practice:

```
ST ~/dev/cli/smartthings-cli $ smartthings installedapps:rename
no installed apps found
```
vs.
```
ST ~/dev/cli/smartthings-cli $ smartthings installedapps:rename
no items found
```
To me just using "items" might be more consistent with other places we do that but it's inconsistent with making the CLI easier to use. Contrast this to what we do when there are items found...we display them and let the user pick one.

I can remove it if you really have strong feelings but I do think it's helpful.

PR #129, opened 2020-12-15
test: split up and add tests for code in basic-io.ts

rossiam:
* pulled some code from basic-io.ts out into its own file called format.ts
* wrote unit tests for all code in basic-io.ts and format.ts
* moved some shared code for unit tests into a "lib" directory


packages/lib/src/__tests__/basic-io.test.ts
```diff
@@ -0,0 +1,189 @@
...
+import { buildMockCommand } from './test-lib/mock-command'
+import { SimpleType } from './test-lib/simple-type'
+
+
+describe('basic-io', () => {
+	const item = { 'str': 'string', num: 5 }
+	const list = [item]
+	let baseCommand: SmartThingsCommandInterface
+	let formatAndWriteItemSpy: jest.SpyInstance<Promise<void>, [command: (SmartThingsCommandInterface & format.TableCommonOutputProducer<unknown>) | (SmartThingsCommandInterface & format.CustomCommonOutputProducer<unknown>), item: unknown, defaultIOFormat?: IOFormat | undefined]>
+	let formatAndWriteListSpy: jest.SpyInstance<Promise<void>, [command: (SmartThingsCommandInterface & Sorting & Naming) | (SmartThingsCommandInterface & format.TableCommonListOutputProducer<unknown> & Naming) | (SmartThingsCommandInterface & format.CustomCommonListOutputProducer<unknown> & Naming), list: unknown[], includeIndex?: boolean | undefined]>
```
john-u, line 20: These lines are quite verbose. I thought the type safety would be covered by the original function declaration. Or is there some jest functionality that can only be used with these declarations in place?

packages/lib/src/__tests__/basic-io.test.ts
```diff
@@ -0,0 +1,189 @@
...
+			...buildMockCommand(),
+			flags: {
+				output: 'output.yaml',
+			},
+		}
+		formatAndWriteItemSpy = jest.spyOn(format, 'formatAndWriteItem')
+		formatAndWriteItemSpy.mockImplementation(async () => { /* empty */})
+
+		formatAndWriteListSpy = jest.spyOn(format, 'formatAndWriteList')
+		formatAndWriteListSpy.mockImplementation(async () => { /* empty */})
```
john-u, line 33: I believe the spy declarations could be moved out of beforeEach as they should be managed by the call to `clearAllMocks` below.

packages/lib/src/__tests__/basic-io.test.ts
```diff
@@ -0,0 +1,189 @@
...
+			const command = {
+				...baseCommand,
+				listTableFieldDefinitions: [],
+				primaryKeyName: 'num',
+				sortKeyName: 'str',
+
+			}
+
+			const getDataMock = jest.fn()
+			getDataMock.mockResolvedValue(list)
```
john-u, line 69: These can be achieved in line as well
```typescript
const getDataMock = jest.fn().mockResolvedValue(list)
```

packages/lib/src/__tests__/basic-io.test.ts
rossiam: With `let` and no immediate assignment, the compiler has no way of determining the type. I know they're ugly, but VSCode figures them out for you (for spies at least).

packages/lib/src/__tests__/basic-io.test.ts
rossiam: They could be but I put them here to reduce duplication. There is so much duplication in these tests that I haven't really figured out how to clean up properly.

packages/lib/src/__tests__/basic-io.test.ts
rossiam: I've updated to make use of this now in several places, including my spies.

packages/lib/src/__tests__/basic-io.test.ts
john-u: Can we just immediately assign then? I only see assertions on calls below which should be successfully cleared between tests with `jest.clearAllMocks()`

packages/lib/src/__tests__/basic-io.test.ts
john-u: I meant to say, move them up into the `describe` context where they are initially declared. This will achieve the reusability while also leveraging the call to `clearAllMocks` after each test to clear state.

packages/lib/src/__tests__/basic-io.test.ts
rossiam: Apparently I can...I had it in my head that `clearAllMocks` meant that I had to start *completely* over, even with the calling `jest.fn` or `jest.spyOn` but testing proves that not to be the case--I've updated these to assign immediately and use const.

packages/lib/src/__tests__/basic-io.test.ts
rossiam: Well that works out because based on our conversation above, that's what I ended up doing. 😄

packages/lib/src/__tests__/basic-io.test.ts
```diff
@@ -0,0 +1,178 @@
...
+		flags: {
+			output: 'output.yaml',
+		},
+	}
+	const formatAndWriteItemSpy = jest.spyOn(format, 'formatAndWriteItem')
+	const formatAndWriteListSpy = jest.spyOn(format, 'formatAndWriteList')
+
+	beforeEach(() => {
+		formatAndWriteItemSpy.mockImplementation(async () => { /* empty */ })
+		formatAndWriteListSpy.mockImplementation(async () => { /* empty */ })
```
john-u, line 29: Since the tests below do not override with any other implementations, these would also be safe to move up to the point of initialization.

packages/lib/src/__tests__/format.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+	afterEach(() => {
+		jest.clearAllMocks()
+	})
+
+	describe('formatAndWriteItem', () => {
+		let itemTableFormatterSpy: jest.SpyInstance<output.OutputFormatter<unknown>, [tableGenerator: TableGenerator, fieldDefinitions: TableFieldDefinition<unknown>[]]>
+
+		beforeEach(() => {
+			itemTableFormatterSpy = jest.spyOn(output, 'itemTableFormatter')
```
john-u, line 38: We could also move this out of beforeEach to clear up the long type above.

packages/lib/src/__tests__/format.test.ts
```diff
@@ -0,0 +1,224 @@
...
+			expect(commonFormatter(item)).toBe('common output')
+			expect(command.buildTableOutput).toHaveBeenCalledTimes(1)
+		})
+	})
+
+	describe('formatAndWriteList', () => {
+		let listTableFormatterSpy: jest.SpyInstance<output.OutputFormatter<unknown[]>, [tableGenerator: TableGenerator, fieldDefinitions: TableFieldDefinition<unknown>[], includeIndex?: boolean | undefined]>
+
+		beforeEach(() => {
+			listTableFormatterSpy = jest.spyOn(output, 'listTableFormatter')
```
john-u, line 90: This can be moved out of beforeEach to spy initialization as well. I see `listTableFormatterSpy` gets its implementation mocked in select tests, so we would want to add a call to [restore](https://jestjs.io/docs/en/mock-function-api#mockfnmockrestore) it between tests (or call it at the end of the tests that set it).

packages/lib/src/__tests__/format.test.ts
rossiam: Sorry, I thought I had got them all. Fixed.

packages/lib/src/__tests__/format.test.ts
rossiam: Fixed

john-u (approved):

packages/lib/src/__tests__/basic-io.test.ts
rossiam: Sorry, I did not mean to ignore this. I'm gonna merge anyway and fix in another pull request.

PR #134, opened 2021-01-06
fix: fix rules lookup without location id and functional refactor

rossiam:
* Updated several `ListingOutputAPICommand` commands to use the new `outputListing` functional style.
    * Notably un-updated is the `schema` command. I need to figure out how to create a schema app so I can test it first.
    * The `config` command no longer has a dependency on `APICommand` which is good since it's not an API command. 😄 
* Fixed `rules` command when looking up a rule by id or index without specifying a location id.
* Commands now use `build(List)TableOutput` before checking the `(listT|t)ableFieldDefinitions`. I didn't actually end up using this but they should be checked in this order since a complex `build(List)TableOutput` method could make use of table field definitions and there is no reason we shouldn't be able to use the default names.
* Moved logic for determining output format into a function so we could use it in the `config` command. (This made tests nicer too.)
* Updated unit tests for library refactors.
* Refactored rules commands to remove duplication.
* Fixed camel case `locationId` in rules command to be kebab case.
* Marked `location-id` flag in `rules:create` command required since it is. We should make it not required and ask the user for the location if it's not specified at some point.

### DIFF
--- a/packages/cli/src/commands/rules.ts
+++ b/packages/cli/src/commands/rules.ts
@@ -1,20 +1,21 @@
 import { flags } from '@oclif/command'
-import { LocationItem, Rule } from '@smartthings/core-sdk'
+import { CLIError } from '@oclif/errors'
 
-import { APICommand, ListingOutputAPICommand, TableFieldDefinition } from '@smartthings/cli-lib'
+import { LocationItem, Rule, SmartThingsClient } from '@smartthings/core-sdk'
+
+import { APICommand, outputListing, TableFieldDefinition } from '@smartthings/cli-lib'
 
 
 export const tableFieldDefinitions: TableFieldDefinition<Rule>[] = ['name', 'id',
 	{ label: 'Num Actions', value: rule => rule.actions.length.toString() },
 	'timeZoneId']
 
-export async function getRulesByLocation(this: APICommand, locationId?: string): Promise<RuleWithLocation[]> {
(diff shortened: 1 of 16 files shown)


packages/cli/src/commands/rules.ts
```diff
@@ -55,17 +57,26 @@ export default class RulesCommand extends ListingOutputAPICommand<Rule, RuleWith
...
+				if (flags.locationId) {
+					return this.client.rules.get(id, flags.locationId)
+				}
+				const allRules = await this.getRulesByLocation(flags.locationId)
+				const rule = allRules.find(rule => rule.id === id)
+				if (!rule) {
+					throw new CLIError(`could not find rule with id ${id}`)
+				}
+				return rule
+			},
```
john-u, line 79: It would be a good refactor to extract this closure to a method/function like `getRulesByLocation` is.

john-u (approved):

packages/cli/src/commands/rules.ts
rossiam: Good point...and it should be used in the `update` command which is almost certainly broken as well. I'm gonna make this pull request a draft and fix that too.

rossiam: (I guess I didn't remove roomId...I can do that though.)

john-u (approved):

PR #136, opened 2021-01-06
fix: correct ordering of InputProcessor calls

rossiam:
* call `InputProcessor` methods in the correct order
* added `jest-extended` (for `toHaveBeenCalledBefore` test)
* updated tests

### DIFF
--- a/packages/lib/src/basic-io.ts
+++ b/packages/lib/src/basic-io.ts
@@ -49,12 +49,13 @@ export async function inputAndOutputItem<I, O>(command: SmartThingsCommandInterf
 		executeCommand: ActionFunction<I, O>) : Promise<void> {
 	const inputProcessor = buildInputProcessor<I>(command)
 	if (inputProcessor.hasInput()) {
+		const inputItem = await inputProcessor.read()
 		const defaultIOFormat = inputProcessor.ioFormat
 		if (command.flags['dry-run']) {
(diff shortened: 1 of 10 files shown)


packages/lib/src/basic-io.ts
```diff
@@ -49,12 +49,13 @@ export async function inputAndOutputItem<I, O>(command: SmartThingsCommandInterf
...
 	if (inputProcessor.hasInput()) {
+		const inputItem = await inputProcessor.read()
 		const defaultIOFormat = inputProcessor.ioFormat
 		if (command.flags['dry-run']) {
 			const outputFormatter = buildOutputFormatter(command, defaultIOFormat)
-			await writeOutput(outputFormatter(await inputProcessor.read()), command.flags.output)
+			await writeOutput(outputFormatter(inputItem), command.flags.output)
 		} else {
-			const item = await executeCommand(await inputProcessor.read())
+			const item = await executeCommand(inputItem)
```
john-u, line 58: I think this is a good change even just for readability, but do you know why this was executing in the wrong order as is? Was it that the inner await passed as parameter wasn't pausing execution of `inputAndOutputItem`? I wouldn't expect it to work that way.

john-u (approved):

packages/lib/src/basic-io.ts
rossiam: At least one of the input processors doesn't know the format for sure until the data has been read.

PR #139, opened 2021-01-11
fix: Show rule ID in list table

bflorian:
Fixes minor bug where rule ID is not being displayed in the list table


rossiam:
I already have this exact change in another pull request that also fixes looking up by index or id without a location id:

https://github.com/SmartThingsCommunity/smartthings-cli/pull/134/files

It's currently in draft because I'm pretty sure the update and delete commands have the same issue I found with the lookup one.

bflorian:
That's great! Let's just close this in favor of that change. Being able to reference with just an index will be nice.

bflorian:
Closing since issue was fixed in #134

PR #148, opened 2021-01-26
refactor: replace SelectingApiCommand with functions

john-u:
## Types of Changes

### DIFF
--- a/packages/cli/src/commands/locations/rooms/delete.ts
+++ b/packages/cli/src/commands/locations/rooms/delete.ts
@@ -1,15 +1,13 @@
 import { flags } from '@oclif/command'
-import { Room } from '@smartthings/core-sdk'
-
 import { getRoomsByLocation } from '../rooms'
-import { SelectingAPICommand } from '@smartthings/cli-lib'
+import { APICommand, selectAndActOn } from '@smartthings/cli-lib'
 
 
-export default class RoomsDeleteCommand extends SelectingAPICommand<Room> {
+export default class RoomsDeleteCommand extends APICommand {
 	static description = 'delete a room'
 
 	static flags = {
-		...SelectingAPICommand.flags,
+		...APICommand.flags,
 		'location-id': flags.string({
 			char: 'l',
 			description: 'a specific location to query',
@@ -34,7 +32,7 @@ export default class RoomsDeleteCommand extends SelectingAPICommand<Room> {
 		await super.setup(args, argv, flags)
 
 		const roomsPromise = this.getRoomsByLocation(flags['location-id'])
-		await this.processNormally(
+		await selectAndActOn(this,
 			args.id,
 			() => roomsPromise,
 			async id => {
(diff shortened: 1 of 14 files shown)


packages/lib/src/acting-io.ts
```diff
@@ -42,9 +42,14 @@ export async function selectAndActOnGeneric<ID, O, L>(command: SelectingCommand<
...
 	const updatedItem = await actionFunction(chosenId)
+
+	if (successMessage) {
+		process.stdout.write(`${successMessage.replace('{{id}}', JSON.stringify(chosenId))}\n`)
+	}
+
 	return [chosenId, updatedItem]
 }
-selectActOnAndOutputGeneric.flags = outputList.flags
+selectAndActOnGeneric.flags = outputList.flags
```
rossiam, line 52: I don't think this should be here. The flags from `outputList.flags` all pertain to complex output that can be formatted as JSON or YAML which does not exist here.

john-u: I thought this was a typo since there is two occurrences of `selectActOnAndOutputGeneric.flags = outputList.flags` in this module and the other is directly under `selectActOnAndOutputGeneric`.

I have gone through and removed setting flags for this and `selectAndActOn` and instead use `APICommand.flags` where necessary.

rossiam: Oops...yeah, that shouldn't have been there at all. It's nice that I can find my bugs in your reviews so I can blame you for them. :-)

packages/cli/src/commands/locations/rooms/delete.ts
```diff
@@ -33,12 +31,12 @@ export default class RoomsDeleteCommand extends SelectingAPICommand<Room> {
 		const { args, argv, flags } = this.parse(RoomsDeleteCommand)
 		await super.setup(args, argv, flags)
 
-		const roomsPromise = this.getRoomsByLocation(flags['location-id'])
```
rossiam, line 36: I actually like the old name better here since this variable is only a promise of rooms, not a list of actual rooms. That said, I also don't see any reason we couldn't move the `await` to this line since we call `await` every other place we reference it.

packages/lib/src/acting-io.ts
```diff
@@ -42,9 +42,13 @@ export async function selectAndActOnGeneric<ID, O, L>(command: SelectingCommand<
 		chosenId = await getIdFromUser(command, list)
 	}
 	const updatedItem = await actionFunction(chosenId)
+
+	if (successMessage) {
+		process.stdout.write(`${successMessage.replace('{{id}}', JSON.stringify(chosenId))}\n`)
```
rossiam, line 47: Github won't let me comment on the code I think you should change since you haven't touched it but you could actually remove this line from the `selectAndActOn` method below and instead pass its `successMessage` parameter on to this function.

rossiam (approved):

rossiam (approved):

PR #149, opened 2021-01-27
feat: Added commands to list, show, and delete installed ST schema instances

bflorian:
Added commands to list, show, and delete installed ST schema connector instances. Also updated installed app delete and rename commands to accept location and verbose flags since location is important context for identifying instances that
often have the same names. 

## Types of Changes

### DIFF
--- a/packages/cli/src/commands/installedschema/delete.ts
+++ b/packages/cli/src/commands/installedschema/delete.ts
@@ -1,0 +1,47 @@
+import { flags } from '@oclif/command'
+
+import { InstalledSchemaApp } from '@smartthings/core-sdk'
+
+import {selectAndActOn, APICommand} from '@smartthings/cli-lib'
+
+import { installedSchemaInstances } from '../installedschema'
+
+
+export default class InstalledSchemaAppDeleteCommand extends APICommand {
+	static description = 'delete the installed schema connector instance'
+
+	static flags = {
+		...APICommand.flags,
+		'location-id': flags.string({
+			char: 'l',
+			description: 'filter results by location',
+			multiple: true,
+		}),
+		verbose: flags.boolean({
+			description: 'include location name in output',
+			char: 'v',
+		}),
+	}
+	static args = [{
+		name: 'id',
+		description: 'installed schema connector UUID',
+	}]
+
+	primaryKeyName = 'isaId'
+	sortKeyName = 'appName'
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/installedschema/delete.ts
```diff
@@ -0,0 +1,66 @@
...
+			multiple: true,
+		}),
+		verbose: flags.boolean({
+			description: 'include location name in output',
+			char: 'v',
+		}),
+	}
+	static args = [{
+		name: 'id',
+		description: 'installed schema conntector UUID',
```
rossiam, line 27: conntector -> connector

packages/cli/src/commands/installedschema/delete.ts
```diff
@@ -0,0 +1,66 @@
...
+					} catch(e) {
+						return []
+					}
+				})))
+
+				if (this.flags.verbose) {
+					return await withLocations(this.client, isas)
+				}
+				return isas
+			},
```
rossiam, line 62: This function is the same as the one in `installedschema.ts`. I would make it an exported function there and use the same one here.

packages/cli/src/commands/installedschema/delete.ts
```diff
@@ -0,0 +1,66 @@
...
+		}),
+	}
+	static args = [{
+		name: 'id',
+		description: 'installed schema conntector UUID',
+	}]
+
+	primaryKeyName = 'isaId'
+	sortKeyName = 'appName'
+	protected listTableFieldDefinitions = ['appName', 'partnerName', 'partnerSTConnection', 'isaId']
```
rossiam, line 32: This really shouldn't be marked `protected`.  It is working but I'm not 100% sure how...I'm guessing it's because the `protected` is removed in the conversion to JavaScript. (The compiler is okay with it being protected because it sees the `sortKeyName` and `primaryKeyName` fields which it _thinks_ will be used. Then, at runtime the code sees this (because protected is gone) and uses it. That's my theory anyway.

rossiam (changes requested):
The spelling error is the important thing to fix.

packages/cli/src/commands/installedschema/delete.ts
bflorian: Fixed

packages/cli/src/commands/installedschema/delete.ts
bflorian: Removed protected

PR #157, opened 2021-02-05
test(LocationsCommand): add inital unit tests

john-u:
## Types of Changes

### DIFF
--- a/packages/cli/src/__tests__/commands/locations.test.ts
+++ b/packages/cli/src/__tests__/commands/locations.test.ts
@@ -1,0 +1,36 @@
+import { v4 as uuidv4 } from 'uuid'
+import { outputListing } from '@smartthings/cli-lib'
+import LocationsCommand from '../../commands/locations'
+
+
+jest.mock('@smartthings/cli-lib', () => {
+	const originalLib = jest.requireActual('@smartthings/cli-lib')
+
+	return {
+		...originalLib,
+		outputListing: jest.fn(),
+	}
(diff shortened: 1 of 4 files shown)


packages/cli/src/__tests__/commands/locations.test.ts
```diff
@@ -0,0 +1,36 @@
...
+	const originalLib = jest.requireActual('@smartthings/cli-lib')
+
+	return {
+		...originalLib,
+		outputListing: jest.fn(),
+	}
+})
+
+describe('LocationsCommand', () => {
+	const mockListing = outputListing as unknown as jest.Mock<typeof outputListing>
```
john-u, line 16: This seems required with Typescript in order to use anything useful in the jest mock API.

rossiam (approved):

packages/cli/src/__tests__/commands/locations.test.ts
rossiam: You could probably do something like this:

```
const outputListingMock = jest.fn()
jest.mock('@smarthings/cli-lib', () => {
...
  return {
    ...originalLib,
    outputListing: outputListingMock,
  }
})
```
And then use `outputListingMock` instead.

john-u: I tried that first and ran into [this issue](https://github.com/facebook/jest/issues/10996). It may be a temporary bug.

PR #159, opened 2021-02-11
refactor: convert selectAndActOn to select and then perform action

rossiam:
Refactored all uses of the "select and act on" methods to do the selection and action as separate calls and then removed the methods and their tests. Also, renamed "acting-io.ts" to "select.ts" since that's all it does now.

## Types of Changes

- [X] I have added tests to cover my changes


packages/cli/src/commands/apps/delete.ts
```diff
@@ -11,16 +11,17 @@ export default class AppDeleteCommand extends APICommand {
...
-	sortKeyName = 'displayName'
-
 	async run(): Promise<void> {
 		const { args, argv, flags } = this.parse(AppDeleteCommand)
 		await super.setup(args, argv, flags)
 
-		await selectAndActOn(this, args.id,
-			async () => await this.client.apps.list(),
-			async id => { await this.client.apps.delete(id) },
-			'app {{id}} deleted')
```
john-u, line 24: I think a majority of these output messages in commands were already lowercase. We may want to double check that we aren't introducing inconsistency by only changing a subset of them here.

john-u (approved):

PR #163, opened 2021-02-25
refactor: convert commands using SelectingOutputAPICommand to functional API

rossiam:
* converted commands that used `SelectingOutputAPICommand` to use the new functional API and removed it.
* updated several commands that needed device selection to use shared `chooseDevice` function
* updated several commands that needed device profile selection to use shared `chooseDeviceProfile` function

## Types of Changes

- [X] Any required documentation has been added

### DIFF
--- a/packages/cli/src/commands/deviceprofiles/view.ts
+++ b/packages/cli/src/commands/deviceprofiles/view.ts
@@ -1,9 +1,6 @@
 import { DeviceProfile, DeviceProfileRequest, PresentationDeviceConfigEntry } from '@smartthings/core-sdk'
 
-import {
-	APICommand,
-	SelectingOutputAPICommand,
-} from '@smartthings/cli-lib'
+import { APICommand, ListingOutputConfig, outputListing, TableGenerator } from '@smartthings/cli-lib'
 
 
 export interface DeviceView {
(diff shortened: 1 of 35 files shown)


packages/lib/src/select.ts
```diff
@@ -4,32 +4,48 @@ import { CommonListOutputProducer } from './format'
...
+		getIdFromUser: IdRetrievalFunction<ID, L>, promptMessage?: string, autoChoose = false): Promise<ID> {
 	if (preselectedId) {
 		return preselectedId
 	}
 
-	const list = await outputList(command, config, listItems, true)
+	const items = await listItems()
+	if (autoChoose && items.length === 1) {
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
```
john-u, line 25: It would be helpful to mention what we know we are ignoring in the comment.

rossiam: Updated ignore comment.

packages/cli/src/commands/deviceprofiles/view.ts
```diff
@@ -134,45 +131,45 @@ export function augmentPresentationValues(view: DeviceView): DeviceView {
...
+
+		const getDeviceProfileAndConfig = async (id: string): Promise<DeviceDefinition> => {
+			const profile = await this.client.deviceProfiles.get(id)
+			if (profile.metadata) {
+				try {
+					const view = await this.client.presentation.get(profile.metadata.vid, profile.metadata.mnmn)
+					prunePresentationValues(view)
+					return { ...profile, view }
+				} catch (error) {
 					return profile
```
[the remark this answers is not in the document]
rossiam: @bflorian do you remember why this is here? I can't reproduce it so I'm tempted to just remove the `try/catch` altogether but maybe there is a situation where an exception could be thrown? For now, I've added `warn` level logging which will make it show up in the log by default but displays nothing to the user.

john-u (approved):

PR #168, opened 2021-03-18
tests: add unit tests for login-authenticator

rossiam:
* added unit tests for `LoginAuthenticator` class
* fixed a couple of minor issues found

## Types of Changes

- [X] I have added tests to cover my changes

### DIFF
--- a/packages/lib/src/__tests__/login-authenticator.test.ts
+++ b/packages/lib/src/__tests__/login-authenticator.test.ts
@@ -1,0 +1,475 @@
+import fs, { NoParamCallback, PathLike } from 'fs'
+
+import axios, { AxiosRequestConfig, AxiosResponse } from 'axios'
+import getPort from 'get-port'
+import open from 'open'
+
+import { logManager } from '../logger'
+import { LoginAuthenticator } from '../login-authenticator'
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const recording = require('log4js/lib/appenders/recording')
+
+
+jest.mock('fs', () => {
+	const originalLib = jest.requireActual('fs')
+
+	return {
+		...originalLib,
+		mkdirSync: jest.fn(),
+		readFileSync: jest.fn(),
+		writeFileSync: jest.fn(),
+		chmod: jest.fn(),
+	}
+})
+
+const mockApp = {
+	get: jest.fn(),
+	listen: jest.fn(),
+}
+jest.mock('express', () => {
+	return () => mockApp
+})
+
+jest.mock('get-port')
+jest.mock('open')
+jest.mock('axios')
+
+async function delay(milliseconds: number): Promise<void> {
+	return new Promise(resolve => setTimeout(resolve, milliseconds))
+}
+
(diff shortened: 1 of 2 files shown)


packages/lib/src/__tests__/login-authenticator.test.ts
```diff
@@ -0,0 +1,475 @@
...
+
+	function setupAuthenticator(): LoginAuthenticator {
+		LoginAuthenticator.init(credentialsFilename)
+		return new LoginAuthenticator(profileName, clientIdProvider)
+	}
+
+	type ExpressRouteHandler = (request: Request, response: Response) => void
+	const finishHappy = (finishHandler: ExpressRouteHandler): void =>
+		finishHandler(mockFinishRequest as unknown as Request, mockFinishResponse as unknown as Response)
+	// Call handlers like would happen if the user were following the flow in the browser.
```
john-u, line 134: I think this feels too much like an integration test rather than unit tests with the delays in here. I see a couple options to remove the need for this imitation and still get our code tested.

1. Extract the arrow functions out of `login()` so that they can be tested separately. Then login tests can just assert that the correct functions were called the expected times.
2. Alternatively, we could start a separate test collection for "integration" and add the tests using `imitateBrowser` to that so that there aren't delays in the unit tests. CI could run them separately as well.

packages/lib/src/__tests__/login-authenticator.test.ts
```diff
@@ -0,0 +1,475 @@
...
+
+		it('sets _credentialsFile properly', function() {
+			LoginAuthenticator.init(credentialsFilename)
+
+			expect((global as { _credentialsFile?: string })._credentialsFile).toBe(credentialsFilename)
+		})
+	})
+
+	describe('constructor', () => {
+		it('throws exception when not init not called', function() {
```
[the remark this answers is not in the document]
rossiam: fixed

packages/lib/src/__tests__/login-authenticator.test.ts
rossiam: The delays are only to let the promise finish...they are very quick. They are unit tests, not integration tests...they don't test anything outside this class. The `imitateBrowser` method is what allows them to be unit tests...it's just calling the methods that would get called via the browser if it were an integration test or running for real.

rossiam: Actually not to let the promise finish...an `await` would do that. I needed to let enough of the promise finish to be able to get access to `express.get` calls so I could feign what the browser would do before the whole promise could finish.

john-u: Yes since the express server is mocked, it is a unit test. I just thought the name `imitateBrowser` along with the requirement to delay felt off.

Generally I don't think static delays in our unit tests are a good idea even if they are quick. We can't really guarantee that something won't change to slightly increase the required time and cause flakey behavior. I will approve since this works and gets this under test, but still recommend a refactor to remove the delays.

john-u (approved):

PR #174, opened 2021-04-16
feat: accept eventsource init dict

john-u:
* include `@types/eventsource` in dependencies so that consuming projects do not need to install separately
* fix code verifier regex in unit test for login authenticator

## Types of Changes

### DIFF
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -24,6 +24,7 @@
 	"dependencies": {
 		"@oclif/errors": "1.3.3",
 		"@smartthings/core-sdk": "^1.3.3",
+		"@types/eventsource": "^1.1.5",
 		"axios": "^0.21.1",
 		"chalk": "^4.1.0",
 		"cli-table": "^0.3.1",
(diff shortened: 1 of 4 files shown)


packages/lib/package.json
```diff
@@ -24,6 +24,7 @@
 	"dependencies": {
 		"@oclif/errors": "1.3.3",
 		"@smartthings/core-sdk": "^1.3.3",
+		"@types/eventsource": "^1.1.5",
```
rossiam, line 27: Maybe I'm missing something but I don't see any reason for this to be a runtime dependency.

rossiam (approved):

packages/lib/package.json
john-u: It needs to be included in dependencies so that it is provided for consuming projects. Otherwise, those projects would see compilation errors even if they aren't explicitly using the SSE stuff.

john-u: I thought of that, but since this affects projects even if they do not use/care about the SSE related functionality, I didn't want to push that responsibility down to consumers.

PR #180, opened 2021-04-28
feat: add logout command which just removes credential for now

rossiam:
Added logout command which currently just removes local credentials. (This will allow the user to log back in again and get updated credentials.)

## Types of Changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Any required documentation has been added
- [] I have added tests to cover my changes


packages/cli/src/commands/logout.ts
```diff
@@ -0,0 +1,19 @@
...
+	async run(): Promise<void> {
+		const { args, argv, flags } = this.parse(LogoutCommand)
+		await super.setup(args, argv, flags)
+
+		if (this.token) {
+			this.log(`Profile ${this.profileName} is set up using a PAT.`)
+			this.exit()
+		}
+
+		this.authenticator.logout && this.authenticator.logout()
```
john-u, line 16: We should either `await` here or make logout sync.

john-u (approved):

packages/cli/src/commands/logout.ts
rossiam: Fixed

PR #191, opened 2021-06-21
fix: Support standard capabilities in interactive profile creation

bflorian:
Corrects a bug in the current implementation that only allows for the selection of custom capabilities in the interactive `deviceprofiles:create` process. That was because a list of only the custom capabilities was displayed, and the entry was validated by checking for membership in the list. Two legitimate scenarios were therefore unsupported:
1. Standard capabilities (the most common case)
2. Custom capabilities that weren't owned by the user (less common but still valid)

Since there are over 200 standard capabilities, prompting with a list of all of them plus any custom capabilities seems needlessly noisy. On the other hand, being able to select from a list is useful, especially for new users. Therefore this change:
* Doesn't initially prompt with a list
* Prompts the user to enter a `?` symbol if a list is desired
* Prompts with a list of all standard and custom capabilities if `?` is entered
* Also accepts a string followed by `?`, for example `temperature?`
* Prompts with a list of all standard and custom capabilities that contain the string

For example:
```
~$ smartthings deviceprofiles:create
? Device Profile Name: Demo Profile
? Primary capability ID (type ? for a list): switch
? Select an action... Add another capability to this component
? Capability ID (type ? for a list): temperature?
┌───┬────────────────────────────┬─────────┬────────┐
│ # │ Id                         │ Version │ Status │
├───┼────────────────────────────┼─────────┼────────┤
│ 1 │ colorTemperature           │ 1       │ live   │
│ 2 │ statelessTemperatureButton │ 1       │ live   │
│ 3 │ temperatureAlarm           │ 1       │ live   │
│ 4 │ temperatureMeasurement     │ 1       │ live   │
└───┴────────────────────────────┴─────────┴────────┘
? Capability ID: 4
? Select an action... Finish & Create
Basic Information
┌───────────────────┬──────────────────────────────────────┐
│ Name              │ Demo Profile                         │
│ main component    │ switch                               │
│                   │ temperatureMeasurement               │
│ Id                │ fbd8626b-e7ba-4d51-a779-81f6b85441f4 │
│ Device Type       │                                      │
│ OCF Device Type   │                                      │
│ Manufacturer Name │ SmartThingsCommunity                 │
│ Presentation ID   │ 5f5f5814-d577-3738-87f5-b03314cab296 │
│ Status            │ DEVELOPMENT                          │
└───────────────────┴──────────────────────────────────────┘
```

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/cli/src/commands/capabilities.ts
+++ b/packages/cli/src/commands/capabilities.ts
@@ -164,6 +164,15 @@ export async function getStandard(client: SmartThingsClient): Promise<Capability
 	return caps.map((capability: CapabilitySummary) => { return { ...capability, namespace: 'st' } })
 }
 
+export async function getAllFiltered(client: SmartThingsClient, filter: string): Promise<CapabilitySummaryWithNamespace[]> {
+	const list = (await Promise.all([getStandard(client), getCustomByNamespace(client)])).flat()
+	if (filter) {
+		filter = filter.toLowerCase()
+		return list.filter(capability => capability.id.toLowerCase().includes(filter) && capability.status !== 'deprecated')
+	}
+	return list
+}
+
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 export async function getIdFromUser(fieldInfo: Sorting, list: CapabilitySummaryWithNamespace[], promptMessage?: string): Promise<CapabilityId> {
 	const convertToId = (itemIdOrIndex: string): string | false => {
@@ -245,6 +254,16 @@ export async function chooseCapability(command: APICommand, idFromArgs?: string,
 		listTableFieldDefinitions: ['id', 'version', 'status'],
 	}
 	return selectGeneric(command, config, preselectedId, () => getCustomByNamespace(command.client), getIdFromUser, prompt)
+}
+
+export async function chooseCapabilityFiltered(command: APICommand, prompt: string, filter: string): Promise<CapabilityId> {
+	const config = {
+		itemName: 'capability',
+		primaryKeyName: 'id',
+		sortKeyName: 'id',
+		listTableFieldDefinitions: ['id', 'version', 'status'],
+	}
+	return selectGeneric(command, config, undefined, () => getAllFiltered(command.client, filter), getIdFromUser, prompt, false)
 }
 
 export default class CapabilitiesCommand extends APICommand {
--- a/packages/cli/src/commands/deviceprofiles/create.ts
+++ b/packages/cli/src/commands/deviceprofiles/create.ts
@@ -12,7 +12,7 @@ import { APICommand, inputAndOutputItem, userInputProcessor } from '@smartthings
 
 import { buildTableOutput } from '../deviceprofiles'
 import { DeviceDefinitionRequest } from './view'
-import { chooseCapability } from '../capabilities'
+import { CapabilityId, chooseCapabilityFiltered } from '../capabilities'
 
 
 const capabilitiesWithoutPresentations = ['healthCheck', 'execute']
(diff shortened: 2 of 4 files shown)


packages/cli/src/commands/schema/authorize.ts
```diff
@@ -4,6 +4,8 @@ import { addPermission } from '../../lib/aws-utils'
 import { lambdaAuthFlags } from '../../lib/common-flags'
 
 
+export const SCHEMA_AWS_PRINCIPAL = '148790070172'
```
rossiam, line 7: I don't personally use this format for constants in JS and I don't see it much either. The distinction used for it in the Java world just doesn't make as much sense in the JS world. Google's Java code style doc uses this format for "constants" which they then go on to define as "static final fields", which exist in JS but are rare.

https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

packages/cli/src/commands/capabilities.ts
```diff
@@ -164,6 +164,15 @@ export async function getStandard(client: SmartThingsClient): Promise<Capability
 	return caps.map((capability: CapabilitySummary) => { return { ...capability, namespace: 'st' } })
 }
 
+export async function getAllFiltered(client: SmartThingsClient, filter: string): Promise<CapabilitySummaryWithNamespace[]> {
+	const list = (await Promise.all([getStandard(client), getCustomByNamespace(client)])).flat()
+	if (filter) {
+		filter = filter.toLowerCase()
+		return list.filter(it => it.id.toLowerCase().includes(filter) && it.status !== 'deprecated')
+	}
+	return list
```
rossiam, line 173: Should we filter out deprecated ones here too?

packages/cli/src/commands/deviceprofiles/create.ts
```diff
@@ -12,7 +12,7 @@ import { APICommand, inputAndOutputItem, userInputProcessor } from '@smartthings
 
 import { buildTableOutput } from '../deviceprofiles'
 import { DeviceDefinitionRequest } from './view'
-import { chooseCapability } from '../capabilities'
+import {CapabilityId, chooseCapabilityFiltered} from '../capabilities'
```
rossiam, line 15: This is a nit and we should have a linting rule for this (I haven't found the right one yet) but we normally have spaces inside the curly braces here.

packages/cli/src/commands/deviceprofiles/create.ts
```diff
@@ -169,17 +169,47 @@ export default class DeviceProfileCreateCommand extends APICommand {
...
+	protected async capabilityDefined(idStr:string): Promise<boolean> {
+		try {
+			const capability = await this.client.capabilities.get(idStr, 1)
+			return !!capability
+		} catch (e) {
+			return false
+		}
+	}
+	protected async promptAndAddCapability(deviceProfile: DeviceProfileRequest, componentId: string, prompt = 'Capability ID'): Promise<CapabilityId> {
+		let capabilityId: CapabilityId = {id: '', version: 0}
```
rossiam, line 182: We would normally have spaces inside the curly braces here too.

rossiam (approved):

packages/cli/src/commands/schema/authorize.ts
john-u: I think it is a good visual distinction for "hard coded" constants which are intended to remain unchanged, rather than just another `const` which really only can't be reassigned, but could change due to side effects, etc. It is also called out in [Mozilla's docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#basic_const_usage).

packages/cli/src/commands/deviceprofiles/create.ts
john-u: [object-curly-spacing](https://eslint.org/docs/rules/object-curly-spacing) looks right. I also have VSCode set to format changes on save, which helps a lot.

packages/cli/src/commands/schema/authorize.ts
```diff
@@ -35,7 +37,7 @@ export default class SchemaAppAuthorizeCommand extends SmartThingsCommand {
 		const { args, argv, flags } = this.parse(SchemaAppAuthorizeCommand)
 		await super.setup(args, argv, flags)
 
-		const principal = flags.principal || '148790070172'
+		const principal = flags.principal && SCHEMA_AWS_PRINCIPAL
```
john-u, line 40: I believe this was meant to be `??` instead of `&&`.

packages/cli/src/commands/schema/create.ts
```diff
@@ -27,7 +28,7 @@ export default class SchemaAppCreateCommand extends APICommand {
 		const createApp = async (_: void, data: SchemaAppRequest): Promise<SchemaCreateResponse> => {
 			if (flags.authorize) {
 				if (data.hostingType === 'lambda') {
-					const principal = flags.principal || '148790070172'
+					const principal = flags.principal && SCHEMA_AWS_PRINCIPAL
```
john-u, line 31: `??` instead of `&&`?

packages/cli/src/commands/capabilities.ts
```diff
@@ -164,6 +164,15 @@ export async function getStandard(client: SmartThingsClient): Promise<Capability
 	return caps.map((capability: CapabilitySummary) => { return { ...capability, namespace: 'st' } })
 }
 
+export async function getAllFiltered(client: SmartThingsClient, filter: string): Promise<CapabilitySummaryWithNamespace[]> {
+	const list = (await Promise.all([getStandard(client), getCustomByNamespace(client)])).flat()
+	if (filter) {
+		filter = filter.toLowerCase()
+		return list.filter(it => it.id.toLowerCase().includes(filter) && it.status !== 'deprecated')
```
john-u, line 171: I would opt for naming the filter param something like `capability` instead of `it`.

john-u (changes requested):

packages/cli/src/commands/capabilities.ts
bflorian: We allow deprecated capabilities to be used if explicitly entered, but it seems to me like we should not be prompting the user to add them, since we don't really want them to be used in new devices

packages/cli/src/commands/deviceprofiles/create.ts
bflorian: Changed

packages/cli/src/commands/deviceprofiles/create.ts
bflorian: Changed

packages/cli/src/commands/schema/authorize.ts
bflorian: Changed. Good catch

packages/cli/src/commands/schema/create.ts
bflorian: Changed. Good catch

packages/cli/src/commands/capabilities.ts
bflorian: Changed

john-u (approved):

PR #215, opened 2021-09-29
fix: Don't consider lack of custom capabilities an error in list command

bflorian:
The CLI is currently returning an error from the `capabilities` command when the user has no custom capabilities. That's inconsistent with all other list commands that return a "no items found" message.

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)


rossiam (approved):
Removing this will result in the parent function (`getCustomByNamespace`) returning an empty list when there are no namespaces but I looked through all the places that use `getCustomByNamespace` and it looks like they all produce better error messages than this one if they get an empty list.

john-u (approved):

PR #238, opened 2021-12-30
perf: migrate to @oclif/core

john-u:
- [x] decide on [which oclif flags to keep](https://github.com/oclif/core/blob/main/MIGRATION.md#restore--h--v-and-version)


packages/cli/README.md
```diff
@@ -233,1620 +232,2101 @@ get a specific app or a list of apps
 
 ```
 USAGE
-  $ smartthings apps [ID]
+  $ smartthings apps [ID] [-h] [-p <value>] [-t <value>] [--language <value>] [--indent <value>] [-j] [-y]
+    [-o <value>] [--compact] [--expanded] [--type <value>] [--classification <value>] [-v]
```
john-u, line 236: The changes to readme output are from migrating away from `@oclif/dev-cli` to `oclif` instead.

packages/cli/README.md
```diff
@@ -188,7 +188,6 @@ that map to the API spec.
 * [`smartthings edge:drivers:uninstall [DRIVERID]`](#smartthings-edgedriversuninstall-driverid)
 * [`smartthings generate:java`](#smartthings-generatejava)
 * [`smartthings generate:node`](#smartthings-generatenode)
-* [`smartthings help [COMMAND]`](#smartthings-help-command)
```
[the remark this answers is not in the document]
rossiam: I don't think it's necessary.

packages/cli/src/hooks/init/init-config.ts
```diff
@@ -1,4 +1,4 @@
-import { Hook } from '@oclif/config'
+import type { Hook } from '@oclif/core'
```
rossiam, line 1: Was this intentional?

rossiam (approved):

packages/cli/src/hooks/init/init-config.ts
john-u: Yes, I thought I'd change this to a [type-only import](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export).

PR #248, opened 2022-01-25
feat: warn when config extensions are .yml

john-u:
- replace util.promisified fs exports with fs.promises equavalents
- add missed logger stub in unit tests

## Types of Changes


packages/cli/src/__tests__/commands/rules/create.test.ts
```diff
@@ -33,7 +33,7 @@ describe('RulesCreateCommand', () => {
 	it('allows user to choose location', async () => {
 		chooseLocationMock.mockResolvedValue('chosen-location-id')
 
-		await expect(RulesCreateCommand.run()).resolves.not.toThrow()
+		await expect(RulesCreateCommand.run([])).resolves.not.toThrow()
```
rossiam, line 36: Sorry, I keep forgetting those!

packages/lib/src/io-util.ts
```diff
@@ -74,3 +74,31 @@ export function stdinIsTTY(): boolean {
...
+
+/**
+ * Check for existence of YAML file.
+ * Enforces official file extension and warns user if incorrect file detected.
+ */
+export function yamlExists(filepath: string): boolean {
+	const parsedPath = path.parse(filepath)
+
+	if (parsedPath.ext !== '.yaml') {
+		return false
```
rossiam, line 86: Not a big deal but I think I would throw an exception here since this method isn't really valid for a non-yaml file (this would makes `false` truly mean only the file does not exist).

rossiam (approved):

packages/lib/src/io-util.ts
john-u: Good call.

PR #256, opened 2022-02-11
fix(deps): update dependencies

john-u:
- pin log4js@6.3.0 until ESM bug is resolved (see oclif/oclif/issues/695)
- remove `generate` command as it was not functional
- add missing/remove redundant dependencies from lerna packages


packages/lib/src/__tests__/login-authenticator.test.ts
```diff
@@ -132,13 +132,13 @@ describe('LoginAuthenticator', () => {
...
 
-	const mkdirMock = fs.mkdirSync as jest.Mock<typeof fs.mkdirSync>
+	const mkdirMock = jest.mocked(fs.mkdirSync)
 	const readFileMock = (fs.readFileSync as unknown as jest.Mock<Buffer, [fs.PathLike]>)
 		.mockImplementation(() => { throw { code: 'ENOENT' } })
 	const writeFileMock = fs.writeFileSync as jest.Mock<typeof fs.writeFileSync>
 	const chmodMock = fs.chmod as unknown as jest.Mock<void, [PathLike, string | number, NoParamCallback]>
 
-	const getPortMock = (getPort as unknown as jest.Mock<Promise<number>, [getPort.Options | undefined]>)
+	const getPortMock = jest.mocked(getPort)
```
rossiam, line 141: This is awesome...I need to remember to use it from now on!

rossiam (approved):

packages/cli/src/commands/config.ts
```diff
@@ -57,7 +57,7 @@ export default class ConfigCommand extends SmartThingsCommand {
 		]
 		const tableFieldDefinitions: TableFieldDefinition<ConfigItem>[] = [
 			...listTableFieldDefinitions,
-			{ label: 'Definition', value: (item) => yaml.safeDump(item.data) },
+			{ label: 'Definition', value: (item) => yaml.dump(item.data) },
```
john-u, line 60: All of these are equivalently "safe". See [migration-guide-from-js-yaml3-to-js-yaml4](https://github.com/nodeca/js-yaml/blob/master/migrate_v3_to_v4.md#migration-guide-from-js-yaml3-to-js-yaml4)

packages/lib/src/__tests__/login-authenticator.test.ts
john-u: Yeah, it was just added in Jest 27. used to be a hidden util in ts-jest.

PR #268, opened 2022-03-18
refactor: clean up old CLI config

rossiam:
Major refactor of class from `cli-config.ts` into interfaces and functions.

* refactor `CLIConfig` class into interfaces and functions
* added support for saving values to "managed config" file
* removed global `cliConfig` variable and added as a property to `SmartThingsCommand`
* made `headers` a property of `APICommand` so we can use it in edge plugin instead of creating a duplicate
* added `useConfigDefault` to `ChooseOptions` (this is not used yet but will be in an upcoming pull request)
* updated some `jest.Mock` casts to use `jest.mocked`
* misc. test updates

⚠️  IMPORTANT: removing the global `cliConfig` will require release of the CLI libraries and then an update to the edge plugin before we can do another CLI binary release. I'll be opening a draft pull request immediately.

## Checklist


packages/lib/src/cli-config.ts
```diff
@@ -1,78 +1,111 @@
...
+				into[profileName] = {}
+			}
+			for (const [configKey, configValue] of Object.entries(profileData)) {
+				into[profileName][configKey] = configValue
+			}
 		}
-		throw new Error(`bad profile configuration for ${name} in ${this._configFile}`)
+		return into
 	}
+	return mergeInto(mergeInto({}, secondary), preferred)
```
john-u, line 82: Is the nested call necessary?

john-u, line 82: Is the nested call necessary?

packages/lib/src/command-util.ts
```diff
@@ -104,15 +104,25 @@ export async function stringGetIdFromUser<L>(fieldInfo: Sorting, list: L[], prom
 	return inputId
 }
 
+/**
+ * Note that not all functions that use this interface support all options. Check the
+ * `chose<Thing>` method itself.
```
john-u, line 109: `chose<Thing>` makes it seem like a method named `chose` which returns a `Thing`.

john-u, line 109: `chose<Thing>` makes it seem like a method named `chose` which returns a `Thing`.

packages/lib/src/smartthings-command.ts
```diff
@@ -56,24 +84,39 @@ export abstract class SmartThingsCommand extends Command implements SmartThingsC
...
+			if (typeof configValue === 'string') {
+				return configValue
+			}
+			this.logger.warn(`expected string value for config key ${keyName} but got ${typeof this.profile[keyName]}`)
+			return undefined
+		}
+		this.logger.trace(`key ${keyName} not found in ${this.profileName} config`)
+	}
+
+	booleanConfigValue(keyName: string, defaultValue = false): boolean {
```
john-u, line 109: It would be useful if all of the config access functions supported a default fallback value.

john-u, line 109: It would be useful if all of the config access functions supported a default fallback value.

packages/lib/src/smartthings-command.ts
```diff
@@ -124,21 +167,24 @@ export abstract class SmartThingsCommand extends Command implements SmartThingsC
...
-			compact = this.profileConfig.compactTableOutput
-		}
-		if (this.flags.expanded) {
-			compact = false
-		} else if (this.flags.compact) {
-			compact = true
-		}
+		this.cliConfig = await loadConfig({
+			configFilename: `${this.config.configDir}/config.yaml`,
+			managedConfigFilename: `${this.config.configDir}/config-managed.yaml`,
```
john-u, line 173: We should store the managed config in the `cacheDir` instead.

john-u:
I think the logic for merging/comparing the managed and user configs could be simplified by only supporting a single profile (ex. `default`) in the managed config. Since we have control over how this file is written, we could make assumptions that would make certain functions simpler and easier to reason about.

packages/lib/src/basic-io.ts
```diff
@@ -86,6 +86,17 @@ export async function outputList<L>(command: SmartThingsCommandInterface, config
...
+ * This is the main function used in most create and update commands. It parses input and passes it
+ * on to the executeAction function parameter.
+ *
+ * By default input is only accepted from stdin or a file specified by the --input command line
+ * flag as JSON or YAML but alternate input processors can be specified as arguments. They simply
+ * need to implement the InputProcessor<I> interface. The most common use-case for this is to use a
+ * Q&A session with the user to build up the input. Another possible use-case would be building
+ * input from command line flags or arguments. Input processors are called in the order they are
+ * specified and the first one to return data is used.
+ */
```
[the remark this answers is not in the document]
rossiam: 👍

packages/lib/src/cli-config.ts
```diff
@@ -1,78 +1,111 @@
-import fs from 'fs'
-import yaml from 'js-yaml'
-import { yamlExists } from './io-util'
+import cliConfigModule from 'js-yaml'
```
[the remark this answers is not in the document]
rossiam: I'm not sure how that happened. I'm guessing I was trying to rename something else and the cursor wasn't where I thought it was but it was definitely not intentional and I've fixed it now.

packages/lib/src/cli-config.ts
```diff
@@ -1,78 +1,111 @@
...
-		if (retVal === null) {
-			throw new Error('null profile specified. Check config.yaml for errors.')
-		}
-		if (typeof retVal === 'object') {
-			return retVal
+/**
+ * Merge profiles from `preferred` and `secondary`, favoring any config entries from `preferred`
+ * over those in `secondary`.
+ */
+export const mergeProfiles = (preferred: ProfilesByName, secondary: ProfilesByName): ProfilesByName => {
```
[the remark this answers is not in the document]
rossiam: `Object.assign` modifies one of the inputs which is not ideal but more importantly, it's just a simple shallow merge and we need the merge that merges at one level down. I was able to replace the inner loop in `mergeInto` with `Object.assign`.

packages/lib/src/cli-config.ts
rossiam: Not strictly, no, but it makes it so the inputs are unmodified which is always a good practice.

packages/lib/src/smartthings-command.ts
rossiam: I've added support for defaults to the `string` version now.

packages/lib/src/smartthings-command.ts
rossiam: Moved to `cacheDir` instead.

john-u (approved):

PR #273, opened 2022-03-29
test: clean up mock type assertions

john-u:
- remove usage of `uuid`

### DIFF
--- a/packages/testlib/jest.config.js
+++ b/packages/testlib/jest.config.js
@@ -15,4 +15,5 @@ module.exports = {
 	modulePathIgnorePatterns: [
 		'<rootDir>/dist',
 	],
+	passWithNoTests: true
 }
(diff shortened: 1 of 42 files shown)


packages/testlib/jest.config.js
```diff
@@ -15,4 +15,5 @@ module.exports = {
 	modulePathIgnorePatterns: [
 		'<rootDir>/dist',
 	],
+	passWithNoTests: true
```
rossiam, line 18: Did you mean to leave this in?

rossiam (approved):

packages/testlib/jest.config.js
john-u: Yes since I deleted the only test file which was a noop, this is required.

PR #297, opened 2022-04-27
refactor: parse once from base command

john-u:
Implements https://github.com/oclif/oclif.github.io/pull/142

Highlights
- Same performance as before
- Maintains type safety of flags in child commands
- Adds type safety to base-level command flags

## Checklist


packages/cli/src/commands/apps.ts
```diff
@@ -35,27 +35,27 @@ export default class AppsCommand extends APICommand {
...
+			if (this.flags.type) {
+				appListOptions.appType = AppType[this.flags.type as keyof typeof AppType]
 			}
-			if (flags.classification) {
-				appListOptions.classification = flags.classification.map(it => AppClassification[it as keyof typeof AppClassification])
+
+			if (this.flags.classification) {
+				// tradeoff for not parsing in here is that TS cannot infer typeof this.constructor.flags
+				const classifications = this.flags.classification as string[]
+				appListOptions.classification = classifications.map(classification => AppClassification[classification as keyof typeof AppClassification])
```
john-u, line 58: I tried a number of things both here and in out base command, but couldn't get flags to infer the static flags types without explicit casting.

john-u, line 58: I tried a number of things both here and in out base command, but couldn't get flags to infer the static flags types without explicit casting.

packages/cli/src/commands/capabilities.ts
```diff
@@ -131,9 +135,9 @@ export interface CapabilityId {
 
 export type CapabilitySummaryWithNamespace = CapabilitySummary & { namespace: string }
 
-export function buildListTableOutput(this: APIOrganizationCommand, capabilities: CapabilitySummaryWithNamespace[]): string {
-	return this.tableGenerator.buildTableFromList(capabilities, ['id', 'version', 'status'])
-}
+// export function buildListTableOutput(this: APIOrganizationCommand, capabilities: CapabilitySummaryWithNamespace[]): string {
+// 	return this.tableGenerator.buildTableFromList(capabilities, ['id', 'version', 'status'])
+// }
```
rossiam, line 140: I'm betting you meant to remove this completely after compiling to make sure it wasn't used.

packages/lib/src/smartthings-command.ts
```diff
@@ -145,70 +177,37 @@ export abstract class SmartThingsCommand extends Command implements SmartThingsC
...
 		const compact = this.flags.expanded
 			? false
 			: (this.flags.compact ? true : this.booleanConfigValue('compactTableOutput', true))
+
 		this._tableGenerator = new DefaultTableGenerator(compact)
 	}
 
 	/**
-	 * This method should be called when the user has decided to not complete a command.
+	 * @deprecated use Command.exit() instead
```
rossiam, line 208: Why not fix this method to use `Command.exit` instead? It also logs a message for the user which `Command.exit` doesn't do and is something we usually want to do in this case.

rossiam (approved):

packages/cli/src/commands/capabilities.ts
john-u: I sure did. Thanks.

PR #339, opened 2022-06-28
feat: update table output

rossiam:
Update table output:
	- switch to table package which handles international characters properly
	- removed compact / expanded command line options
	- removed compactTableOutput configuration option
	- added group-rows and no-group-rows command line options
	- added groupTableOutputRows configuration option
	- (lib) completely isolated use of dependency to table-generator.ts

## Checklist

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -75,14 +75,12 @@
 		"@smartthings/core-sdk": "^5.0.0",
 		"@smartthings/plugin-cli-edge": "^1.14.1",
 		"aws-sdk": "^2.1144.0",
-		"cli-table": "^0.3.11",
 		"inquirer": "^8.2.4",
 		"js-yaml": "^4.1.0",
 		"log4js": "6.3.0"
 	},
 	"devDependencies": {
 		"@smartthings/cli-testlib": "^1.0.0-beta.4",
-		"@types/cli-table": "^0.3.0",
 		"@types/inquirer": "^8.2.1",
 		"@types/jest": "^28.1.3",
 		"@types/js-yaml": "^4.0.5",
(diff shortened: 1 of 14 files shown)


packages/lib/src/__tests__/table-generator.test.ts
```diff
@@ -49,50 +49,45 @@ function stripEscapeSequences(str: string): string {
...
-function rowCount(table: string): number {
-	return table.split('\n').length - 2
-}
+/**
+ * Clean up string created using backticks. Lines are stripped of ` +|` at the beginning of each
+ * line (allowing indentation) and `|` at the end (allowing for spaces at the end of the line
+ * that are not automatically removed by tooling). If there is a new-line at the beginning of the
+ * string, it is also removed.
+ */
+const fixIndent = (input: string): string => input.replace(/^\s*\|/gm, '').replace(/\|$/gm, '').replace(/^\n/, '')
```
john-u, line 58: It is not clear to me why I would need this if I was adding tests here. Is there something preventing us from formatting the expected strings without a function?

packages/cli/package.json
```diff
@@ -75,14 +75,13 @@
 		"@smartthings/core-sdk": "^4.2.0",
 		"@smartthings/plugin-cli-edge": "^1.14.1",
 		"aws-sdk": "^2.1144.0",
-		"cli-table": "^0.3.11",
 		"inquirer": "^8.2.4",
 		"js-yaml": "^4.1.0",
-		"log4js": "6.3.0"
+		"log4js": "6.3.0",
+		"table": "^6.8.0"
```
john-u, line 81: It doesn't look like this is actually required for the cli package.

john-u (commented):
Awesome changes!

packages/cli/package.json
rossiam: Good catch. Fixed.

packages/lib/src/__tests__/table-generator.test.ts
rossiam: It just makes the tests look nicer. We can use the mutli-line backtick quoting and still indent and not worry about the editor stripping the whitespace from the end of the line.

john-u (approved):

PR #392, opened 2022-09-07
feat: support verbose flag for installedapps and installedschema commands with single item

rossiam:
* added `withLocation` to add location to a single item (matches/uses `withLocations` which does the same for a list)
* support verbose flag for installedapps and installedschema commands with single item
* added tests for installedschema command
* removed `InstalledAppWithLocation` and `InstalledSchemaAppWithLocation` in favor of `WithNamedLocation`

## Checklist

### DIFF
--- a/packages/cli/src/commands/installedschema.ts
+++ b/packages/cli/src/commands/installedschema.ts
@@ -2,10 +2,9 @@ import { Flags } from '@oclif/core'
 
 import { InstalledSchemaApp } from '@smartthings/core-sdk'
 
-import { APICommand, OutputItemOrListConfig, outputItemOrList } from '@smartthings/cli-lib'
+import { APICommand, OutputItemOrListConfig, outputItemOrList, WithNamedLocation, withLocation } from '@smartthings/cli-lib'
 
 import {
-	InstalledSchemaAppWithLocation,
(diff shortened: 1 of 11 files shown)


packages/cli/src/commands/installedschema.ts
```diff
@@ -36,21 +35,25 @@ export default class InstalledSchemaAppsCommand extends APICommand<typeof Instal
...
 			listTableFieldDefinitions,
 			tableFieldDefinitions,
 		}
 
 		if (this.flags.verbose) {
-			config.listTableFieldDefinitions.splice(3, 0, 'location')
+			config.listTableFieldDefinitions = [...listTableFieldDefinitions]
+			config.listTableFieldDefinitions.splice(3, 0, 'locationId', 'location')
+			config.tableFieldDefinitions = [...tableFieldDefinitions]
 			config.tableFieldDefinitions.splice(5, 0, 'location')
```
rossiam, line 49: The reason I'm duplicating the arrays here before splicing is so we don't modify the original array which breaks unit tests. It might be cleaner to change the arrays in the util class to functions that return the array (as a fresh copy each time) but I figured this is good enough for now.

Sitlintac (approved):

PR #398, opened 2022-09-12
fix: remove "-id" suffix from flags for consistency

rossiam:
Removed "-id" suffix from command line flags that use it. Most of these are changing `--location-id` to `--location` but there are a few others mixed in.

## Checklist

### DIFF
--- a/.changeset/early-zoos-speak.md
+++ b/.changeset/early-zoos-speak.md
@@ -1,0 +1,6 @@
+---
+"@smartthings/cli": patch
+"@smartthings/cli-lib": patch
+---
+
+IMPORTANT: removed `-id` suffix from command line flags that had them for consistency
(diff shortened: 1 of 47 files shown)


john-u (approved):
This is good cleanup!

I could see this causing initial confusion depending on the descriptions of each flag if the user didn't know a UUID or name was required. One thing that could help with this is the [helpValue](https://github.com/oclif/core/blob/4c3466bf2399ad1a001302f593785d3689840011/src/interfaces/parser.ts#L155) flag option which could make the flag specifications more clear in help like `location=<UUID>` instead of `location=<value>`.

rossiam:
> This is good cleanup!
> 
> I could see this causing initial confusion depending on the descriptions of each flag if the user didn't know a UUID or name was required. One thing that could help with this is the [helpValue](https://github.com/oclif/core/blob/4c3466bf2399ad1a001302f593785d3689840011/src/interfaces/parser.ts#L155) flag option which could make the flag specifications more clear in help like `location=<UUID>` instead of `location=<value>`.

I just noticed this and I already merged. I like it though so I'll open another pull request and add these.

PR #407, opened 2022-09-19
ci: add ARM build support

john-u:
- replace bash script with Node script
- tar.gz files instead of using zip


.eslintrc.js
```diff
@@ -142,5 +142,6 @@ module.exports = {
 				format: ['PascalCase', 'UPPER_CASE'],
 			},
 		],
+		"no-console": "off",
```
john-u, line 145: From https://eslint.org/docs/latest/rules/no-console#when-not-to-use-it
> If you’re using Node.js, however, console is used to output information to the user and so is not strictly used for debugging purposes. If you are developing for Node.js then you most likely do not want this rule enabled.

rossiam (approved):

PR #416, opened 2022-09-26
chore: update TypeScript

rossiam:
## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
 	"private": true,
 	"//": {
 		"scripts": {
-			"version": "TODO: remove 'npm i --package-lock-only' when this is resolved https://github.com/changesets/changesets/issues/421"
+			"version": "TODO: remove 'npm i --package-lock-only' when this is resolved https://github.com/changesets/changesets/issues/421 "
 		}
 	},
 	"scripts": {
@@ -34,6 +34,8 @@
 		"@commitlint/cli": "^17.1.2",
 		"@commitlint/config-conventional": "^17.1.0",
 		"cz-conventional-changelog": "^3.3.0",
-		"rimraf": "^3.0.2"
+		"rimraf": "^3.0.2",
+		"ts-node": "^10.9.1",
+		"typescript": "^4.8.3"
 	}
 }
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
 	"private": true,
 	"//": {
 		"scripts": {
-			"version": "TODO: remove 'npm i --package-lock-only' when this is resolved https://github.com/changesets/changesets/issues/421"
+			"version": "TODO: remove 'npm i --package-lock-only' when this is resolved https://github.com/changesets/changesets/issues/421 "
```
rossiam, line 6: The space just makes this temporary link work in VSCode.

package.json
```diff
@@ -34,6 +34,8 @@
 		"@commitlint/cli": "^17.1.2",
 		"@commitlint/config-conventional": "^17.1.0",
 		"cz-conventional-changelog": "^3.3.0",
-		"rimraf": "^3.0.2"
+		"rimraf": "^3.0.2",
+		"ts-node": "^10.9.1",
+		"typescript": "^4.8.3"
```
john-u, line 39: Is there a specific reason these two were moved here?

john-u (approved):

package.json
rossiam: Just so we wouldn't have to change it in multiple places in the future. It looked to me like all the other shared `devDependencies` were here. (Lerna had some weird thing where they were needed in the individual projects but I'm not aware of any such limitation for workspaces.)

Is there a reason I'm not aware of for keeping them in all the sub-projects?

john-u: The only devDependencies in root are either only required there (changesets) or are a direct root dependency (rimraf) and are also installed into packages. It depends on preference, but I was generally keeping devDependencies for monorepo management in the root and other dependencies explicitly in each package to allow for flexibility (moving packages in/out of monorepo, partial upgrades on one package that may break another, etc.). I don't see it as an issue right away, but it's something to be aware of with the keeping everything in root approach: if you want to update a specific package but the new version breaks other packages you might not care about for that change, you are required to either split them out then or resolve all upgrade issues together.

PR #426, opened 2022-10-17
Summary info

rossiam:
Removed "this is summarized" text from code and added it to the README.

## Checklist

### DIFF
--- a/packages/cli/src/commands/apps.ts
+++ b/packages/cli/src/commands/apps.ts
@@ -29,6 +29,15 @@ export default class AppsCommand extends APICommand<typeof AppsCommand.flags> {
 		description: 'the app id or number from list',
 	}]
 
+	static examples = [
+		{ description: 'list all apps', command: 'smartthings apps' },
+		{ description: 'list the first app in the list retrieved by running "smartthings apps"', command: 'smartthings apps 1' },
+		{ description: 'list an app by id', command: 'smartthings apps <app-id>' },
+		{ description: 'include URLs and ARNs in the output', command: 'smartthings apps --verbose' },
+		{ description: 'list only SERVICE classification devices', command: 'smartthings apps --classification SERVICE' },
+		{ description: 'list only API only apps', command: 'smartthings apps --type API_ONLY' },
+	]
+
 	async run(): Promise<void> {
 		const config: OutputItemOrListConfig<AppResponse, PagedApp | AppResponse> = {
 			primaryKeyName: 'appId',
--- a/packages/edge/src/lib/commands/drivers-util.ts
+++ b/packages/edge/src/lib/commands/drivers-util.ts
@@ -17,7 +17,6 @@ import {
 	selectFromList,
 	SelectFromListConfig,
 	stringTranslateToId,
-	summarizedText,
 	TableFieldDefinition,
 	TableGenerator,
 } from '@smartthings/cli-lib'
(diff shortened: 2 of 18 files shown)


packages/edge/src/lib/commands/drivers-util.ts
```diff
@@ -44,8 +43,7 @@ export const buildTableOutput = (tableGenerator: TableGenerator, driver: EdgeDri
 	}
 	return `Basic Information\n${basicInfo}\n\n` +
 		`${deviceIntegrationProfiles}\n\n` +
-		`${fingerprints}\n\n` +
-		summarizedText
+		fingerprints
```
Sitlintac, line 46: In similar places in other files, you left the variable in its wrapper in the string literal even though the newlines were removed.  I could see that keeping all the variables in wrappers might make it easier to skim because they all look the same, but also that it's cleaner to do what you did here.  Is one approach preferable to the other?

packages/cli/src/commands/apps/delete.ts
```diff
@@ -12,6 +12,11 @@ export default class AppDeleteCommand extends APICommand<typeof AppDeleteCommand
 		description: 'App profile UUID or number in the list',
 	}]
 
+	static examples = [
+		{ description: 'select app to delete from list', command: 'smartthings apps:delete' },
+		{ description: 'delete a specific app by id', command: 'smartthings <app-id>' },
```
Sitlintac, line 17: Should this be `smartthings apps:delete <app-id>`?

packages/cli/src/commands/apps.ts
```diff
@@ -29,6 +29,15 @@ export default class AppsCommand extends APICommand<typeof AppsCommand.flags> {
 		description: 'the app id or number from list',
 	}]
 
+	static examples = [
+		{ description: 'list all apps', command: 'smartthings apps' },
+		{ description: 'list the first app in the list retrieved by running "smartthings apps"', command: 'smartthings apps 1' },
+		{ description: 'list an app by id', command: 'smartthings apps <app-id>' },
+		{ description: 'include URLs and ARNs in the output', command: 'smartthings apps --verbose' },
+		{ description: 'list only SERVICE classification devices', command: 'smartthings apps --classification SERVICE' },
+		{ description: 'list only API only apps', command: 'smartthings apps --type API_ONLY' },
```
Sitlintac, line 38: The description might be a little easier to parse if it were 'list only API-only apps'

packages/cli/src/commands/apps/delete.ts
rossiam: Yes it should but this wasn't supposed to be in the PR. I've fixed this PR to not include it at all and I'll update in PR #422.

packages/cli/src/commands/apps.ts
rossiam: Good observation. This also wasn't supposed to be in the PR. I've fixed this PR to not include it at all and I'll update in PR #422.

packages/cli/src/commands/apps/delete.ts
rossiam: Updated in PR #422 now.

packages/cli/src/commands/apps.ts
rossiam: Fixed in #422 now.

Sitlintac (approved):

PR #446, opened 2022-11-15
fix: remove profile dependency from reset test

Sitlintac:
fix bug causing test of reset command to fail if SMARTTHINGS_PROFILE was set

## Checklist

### DIFF
--- a/packages/cli/src/__tests__/commands/config/reset.test.ts
+++ b/packages/cli/src/__tests__/commands/config/reset.test.ts
@@ -2,7 +2,7 @@ import inquirer from 'inquirer'
 
 import ConfigResetCommand from '../../../commands/config/reset'
 
-import { resetManagedConfig } from '@smartthings/cli-lib'
+import { resetManagedConfig, SmartThingsCommand } from '@smartthings/cli-lib'
 
 
 jest.mock('inquirer')
@@ -14,6 +14,8 @@ describe('ConfigResetCommand', () => {
 
 	const resetManagedConfigMock = jest.mocked(resetManagedConfig)
 
+	const profileNameSpy = jest.spyOn(SmartThingsCommand.prototype, 'profileName', 'get')
+
 	it('does nothing if user says no to prompt', async () => {
 		promptMock.mockResolvedValueOnce({ confirmed: false })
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/__tests__/commands/config/reset.test.ts
```diff
@@ -14,6 +14,8 @@ describe('ConfigResetCommand', () => {
 
 	const resetManagedConfigMock = jest.mocked(resetManagedConfig)
 
+	const profileName = process.env.SMARTTHINGS_PROFILE ?? 'default'
```
rossiam, line 17: It would be more "unit"y to not even use the environment variable at all but rather to mock it. Probably the simplest way to mock it (that I can think of) would be to spy on the `profileName` `get` method of the `SmartThingsCommand` prototype:

```
const profileNameSpy = jest.spyOn(SmartThingsCommand.prototype, 'profileName', 'get')
profileNameSpy.mockReturnValue('default')
```

Since spies in jest don't reset the same as everything else, you have to call `mockRestore` on the spy when you're done with it so it doesn't interfere with other methods.

rossiam (approved):

packages/cli/src/__tests__/commands/config/reset.test.ts
Sitlintac: Would it work to instead do `profileNameSpy.mockReturnValueOnce('default')`, or does that not work with spies?

.changeset/thin-adults-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smartthings/cli": patch
+---
+
+remove profile dependency from reset test
```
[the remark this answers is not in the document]
Sitlintac: Removed.  I wasn't sure if a changeset was appropriate or not, and figured it was safer to include it than not.

packages/cli/src/__tests__/commands/config/reset.test.ts
Sitlintac: It turns out `mockReturnValueOnce` without a `mockRestore()` does not work, but your solution does.

PR #462, opened 2023-01-12
feat: enable verbose when getting device

Sitlintac:
- enable verbose flag when getting a single device
- add helper method to get location and room names for a single item

## Checklist

### DIFF
--- a/packages/cli/src/__tests__/commands/devices.test.ts
+++ b/packages/cli/src/__tests__/commands/devices.test.ts
@@ -1,7 +1,8 @@
-import { Device, DevicesEndpoint, SmartThingsClient } from '@smartthings/core-sdk'
+import { Device, DevicesEndpoint, DeviceStatus, SmartThingsClient } from '@smartthings/core-sdk'
 
 import {
 	CustomCommonOutputProducer, DefaultTableGenerator, outputItemOrList,
+	withLocationAndRoom,
 	withLocationsAndRooms, WithNamedRoom,
 } from '@smartthings/cli-lib'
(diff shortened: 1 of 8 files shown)


packages/cli/src/__tests__/commands/devices.test.ts
```diff
@@ -346,4 +348,55 @@ describe('DevicesCommand', () => {
...
+			expect.anything(),
+			deviceId,
+			expect.anything(),
+			expect.anything(),
+		)
+
+		const getFunction = outputItemOrListMock.mock.calls[0][4]
+
+		await expect(getFunction(deviceId)).resolves.toBe(deviceWithRoom)
+		expect(getSpy).toBeCalledWith(deviceId, { includeStatus: undefined })
```
rossiam, line 371: Not super important but I think it would be worth it to also expect that each of these have been called exactly once. (Same in next test.)

packages/cli/src/commands/devices.ts
```diff
@@ -64,7 +65,7 @@ export default class DevicesCommand extends APICommand<typeof DevicesCommand.fla
 			multiple: true,
 		}),
 		verbose: Flags.boolean({
-			description: 'include location name in output',
+			description: 'include location and room name in output',
```
rossiam, line 68: Good catch!

packages/cli/src/__tests__/commands/devices.test.ts
Sitlintac: Done.

PR #475, opened 2023-02-01
chore: prefer types over interfaces

rossiam:
* updated rule to require `type` rather than `interface`
* updated code to conform to this rule
* the one exception where we still need `interface` is for a Jest matcher since the way you add your own matchers is to extend the Jest `Matchers` interface

## Checklist

### DIFF
--- a/packages/cli/src/commands/virtualdevices/events.ts
+++ b/packages/cli/src/commands/virtualdevices/events.ts
@@ -35,7 +35,7 @@ function buildTableOutput(tableGenerator: TableGenerator, data: EventInputOutput
 	return table.toString()
 }
 
-interface EventInputOutput {
+type EventInputOutput = {
 	input: DeviceEvent[]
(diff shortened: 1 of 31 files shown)


packages/cli/src/commands/virtualdevices/events.ts
```diff
@@ -35,7 +35,7 @@ function buildTableOutput(tableGenerator: TableGenerator, data: EventInputOutput
 	return table.toString()
 }
 
-interface EventInputOutput {
+export type EventInputOutput = {
```
Sitlintac, line 38: Just want to double-check that this one wants to be exported, since it wasn't before.

Sitlintac (approved):

packages/cli/src/commands/virtualdevices/events.ts
rossiam: Oops, not intentional. Fixed.

PR #498, opened 2023-04-06
feat: add Modes commands

Sitlintac:
Add commands for working with location modes:
- list/get
- create
- update
- delete
- get current
- set current

## Checklist

### DIFF
--- a/packages/cli/src/commands/locations/modes.ts
+++ b/packages/cli/src/commands/locations/modes.ts
@@ -1,0 +1,67 @@
+import { Flags } from '@oclif/core'
+
+import { Mode } from '@smartthings/core-sdk'
+
+import { APICommand, OutputItemOrListConfig, outputItemOrList, WithNamedLocation } from '@smartthings/cli-lib'
+
+import { getModesByLocation, tableFieldDefinitions, tableFieldDefinitionsWithLocationName } from '../../lib/commands/locations/modes-util'
+
+
+export default class ModeCommand extends APICommand<typeof ModeCommand.flags> {
+	static description = 'list modes or get information for a specific mode'
+
+	static flags = {
+		...APICommand.flags,
+		location: Flags.string({
+			char: 'l',
+			description: 'a specific location to query',
+			helpValue: '<UUID>',
+		}),
+		verbose: Flags.boolean({
+			description: 'include location name in output',
+			char: 'v',
+		}),
+		...outputItemOrList.flags,
+	}
+
+	static args = [{
+		name: 'idOrIndex',
+		description: 'mode UUID or index',
+	}]
+
+	static examples = [
+		{ description: 'list all modes in your location(s)', command: 'smartthings locations:modes' },
+		{ description: 'get details of the third mode in the list retrieved by running "smartthings locations:modes"', command: 'smartthings locations:modes 3' },
+		{ description: 'get details of a mode by its id', command: 'smartthings locations:modes 636169e4-8b9f-4438-a941-953b0d617231' },
+		{ description: 'include location name and ID in the output', command: 'smartthings locations:modes --verbose' },
+		{ description: 'list all modes in a particular location', command: 'smartthings locations:modes --location=5dfd6626-ab1d-42da-bb76-90def3153998' },
+	]
+
+	async run(): Promise<void> {
+		const config: OutputItemOrListConfig<Mode & WithNamedLocation> = {
+			primaryKeyName: 'id',
+			sortKeyName: 'label',
+			listTableFieldDefinitions: tableFieldDefinitions,
+			tableFieldDefinitions: tableFieldDefinitions,
+		}
+		if (this.flags.verbose) {
+			config.listTableFieldDefinitions = tableFieldDefinitionsWithLocationName
+			config.tableFieldDefinitions = tableFieldDefinitionsWithLocationName
+		}
+		const modes = await getModesByLocation(this.client, this.flags.location)
+		await outputItemOrList(this, config, this.args.idOrIndex,
+			async () => modes,
+			async id => {
+				const mode = modes.find(mode => mode.id === id)
+				if (!mode) {
+					throw Error(`could not find mode with id ${id}`)
+				}
+				if (this.flags.verbose) {
+					return mode
+				}
+				// If we aren't using `verbose` flag, get the raw mode again
+				// so there aren't extra fields in JSON/YAML output
+				return this.client.modes.get(id, mode.locationId)
+			})
+	}
+}
--- a/packages/cli/src/commands/locations/modes/delete.ts
+++ b/packages/cli/src/commands/locations/modes/delete.ts
@@ -1,0 +1,39 @@
+import { Flags } from '@oclif/core'
+import { APICommand } from '@smartthings/cli-lib'
+import { chooseMode } from '../../../lib/commands/locations/modes-util'
+
+
+export default class ModesDeleteCommand extends APICommand<typeof ModesDeleteCommand.flags> {
+	static description = 'delete a mode'
+
+	static flags = {
+		...APICommand.flags,
+		location: Flags.string({
+			char: 'l',
+			description: 'a specific location to query',
+			helpValue: '<UUID>',
+		}),
+	}
+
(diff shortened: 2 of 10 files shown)


packages/cli/src/lib/commands/locations/modes-util.ts
```diff
@@ -0,0 +1,54 @@
...
+export const tableFieldDefinitions: TableFieldDefinition<Mode>[] = ['label', 'id', 'name' ]
+export const tableFieldDefinitionsWithLocationName: TableFieldDefinition<Mode & WithNamedLocation>[] = ['label', 'id', 'name', 'location', 'locationId' ]
+
+export async function getModesByLocation(client: SmartThingsClient, locationId?: string): Promise<(Mode & WithNamedLocation)[]> {
+	let locations: LocationItem[] = []
+	if (locationId) {
+		locations = [await client.locations.get(locationId)]
+	} else {
+		locations = await client.locations.list()
+	}
```
rossiam, line 18: Nothing at all wrong with this code but you could use a ternary to avoid the `let`:

```
	const locations = locationId
		? [await client.locations.get(locationId)]
		: await client.locations.list()
```

packages/cli/src/commands/locations/modes/getcurrent.ts
```diff
@@ -0,0 +1,54 @@
...
+			config.listTableFieldDefinitions = tableFieldDefinitionsWithLocationName
+			config.tableFieldDefinitions = tableFieldDefinitionsWithLocationName
+		}
+		const location = await this.client.locations.get(await chooseLocation(this, this.flags.location, true))
+		const currentMode: Mode & WithNamedLocation = {
+			...(await this.client.modes.getCurrent(location.locationId)),
+			location: location.name,
+			locationId: location.locationId,
+		}
+		await formatAndWriteItem(this, config, currentMode)
```
rossiam, line 52: For purposes of more accurate JSON and YAML output, I would avoid adding the extra fields if `--verbose` wasn't specified. You can also use the `withLocation` function to look up the location and add the name. The mode endpint is unusual in that it doesn't include the location id so you do have to add that manually. The code might look something like this:

```
		const locationId = await chooseLocation(this, this.flags.location, true)
		const currentMode = await this.client.modes.getCurrent(locationId)
		const mode = this.flags.verbose
			? await withLocation(this.client, { ...currentMode, locationId } )
			: currentMode
		await formatAndWriteItem(this, config, mode)
```

(`withLocation` gets added as an import from `@smarthings/cli-lib`.)

packages/cli/src/commands/locations/modes/setcurrent.ts
```diff
@@ -0,0 +1,34 @@
...
+	}]
+
+	static examples = [
+		{ description: 'select a mode from a list of all modes and set it to be current for its location', command: 'smartthings locations:modes:setcurrent' },
+		{ description: 'select a mode from a list of modes in a specified location and set it to be current', command: 'smartthings locations:modes:setcurrent --location=5dfd6626-ab1d-42da-bb76-90def3153998' },
+		{ description: 'set the specified mode to be current for its location', command: 'smartthings locations:modes:setcurrent 636169e4-8b9f-4438-a941-953b0d617231' },
+	]
+
+	async run(): Promise<void> {
+		const [modeId, locationId] = await chooseMode(this, this.flags.location, this.args.id)
```
rossiam, line 30: Same comment as on `chooseMode` in delete command.

packages/cli/src/commands/locations/modes/update.ts
```diff
@@ -0,0 +1,37 @@
...
+	}]
+
+	static examples = [
+		{ description: 'select a mode from a list of all modes and update it using the data in new-data.json', command: 'smartthings locations:modes:update -i=new-data.json' },
+		{ description: 'select a mode from a list of modes in a specified location and update it using the data in new-data.json', command: 'smartthings locations:modes:update --location=5dfd6626-ab1d-42da-bb76-90def3153998 -i=new-data.json' },
+		{ description: 'update a specified mode using the data in new-data.json', command: 'smartthings locations:modes:update 636169e4-8b9f-4438-a941-953b0d617231 -i=new-data.json' },
+	]
+
+	async run(): Promise<void> {
+		const [modeId, locationId] = await chooseMode(this, this.flags.location, this.args.id)
```
rossiam, line 32: Same comment as on `chooseMode` in delete command.

rossiam (approved):

rossiam:
This pull request does need a changeset.

rossiam (changes requested):
Most of the comments are optional but we do at least need a changeset.

packages/cli/src/lib/commands/locations/modes-util.ts
Sitlintac: Much better, thanks!

packages/cli/src/lib/commands/locations/modes-util.ts
```diff
@@ -0,0 +1,54 @@
...
+		throw new Errors.CLIError('could not find any locations for your account. Perhaps ' +
+			"you haven't created any locations yet.")
+	}
+
+	let modes: (Mode & WithNamedLocation)[] = []
+	for (const location of locations) {
+		const locationModes = await client.modes.list(location.locationId)
+		modes = modes.concat(locationModes.map(mode => { return { ...mode, locationId: location.locationId, location: location.name } }))
+	}
+	return modes
```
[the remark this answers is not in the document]
Sitlintac: I like the second approach better.  Done.

packages/cli/src/commands/locations/modes/delete.ts
```diff
@@ -0,0 +1,34 @@
...
+	}]
+
+	static examples = [
+		{ description: 'select a mode from a list of all modes and delete it', command: 'smartthings locations:modes:delete' },
+		{ description: 'select a mode from a list of modes in a specified location and delete it', command: 'smartthings locations:modes:delete --location=5dfd6626-ab1d-42da-bb76-90def3153998' },
+		{ description: 'delete a specified mode', command: 'smartthings locations:modes:delete 636169e4-8b9f-4438-a941-953b0d617231' },
+	]
+
+	async run(): Promise<void> {
+		const [modeId, locationId] = await chooseMode(this, this.flags.location, this.args.id)
```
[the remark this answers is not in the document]
Sitlintac: I've put the call to `chooseLocation` at the beginning of `chooseMode` because I think we'll always want to call it at that point and it seemed cleaner than other options.  What do you think?

packages/cli/src/commands/locations/modes.ts
```diff
@@ -0,0 +1,62 @@
...
+		}
+		const modes = await getModesByLocation(this.client, this.flags.location)
+		await outputItemOrList(this, config, this.args.idOrIndex,
+			async () => modes,
+			async id => {
+				const mode = modes.find(mode => mode.id === id)
+				if (!mode) {
+					throw Error(`could not find mode with id ${id}`)
+				}
+				return mode
```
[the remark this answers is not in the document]
Sitlintac: Good catch.  I wasn't thinking about JSON/YAML output here so I fell for the "seems useless" misdirect.  Whoops.

packages/cli/src/commands/locations/modes/getcurrent.ts
Sitlintac: Done.

packages/cli/src/commands/locations/modes/setcurrent.ts
Sitlintac: Same response as on `chooseMode` in delete command.

packages/cli/src/commands/locations/modes/update.ts
Sitlintac: Same response as on `chooseMode` in delete command.

packages/cli/src/commands/locations/modes/delete.ts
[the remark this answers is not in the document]
rossiam: I wouldn't have thought of that but it seems like a good idea to me.

rossiam (approved):

PR #510, opened 2023-05-04
fix: correct capabilities:presentation and deviceprofiles:view commands

Sitlintac:
- fix `capabilities:presentation` and `deviceprofiles:view` commands to prompt for user selection if no ID provided
- minor update to utility library for capabilities

## Checklist

### DIFF
--- a/.changeset/violet-moles-think.md
+++ b/.changeset/violet-moles-think.md
@@ -1,0 +1,7 @@
+---
+"@smartthings/cli": patch
+---
+
+fix  commands to prompt for required inputs when not supplied on the command line
+  * `capabilities:presentation` asks for a capability
+  * `deviceprofiles:view` asks for a deviceprofile
--- a/packages/cli/src/commands/capabilities/presentation.ts
+++ b/packages/cli/src/commands/capabilities/presentation.ts
@@ -2,9 +2,9 @@ import { Flags } from '@oclif/core'
 
 import { CapabilityPresentation } from '@smartthings/core-sdk'
 
-import { APIOrganizationCommand, OutputItemOrListConfig, outputItemOrListGeneric, TableGenerator } from '@smartthings/cli-lib'
+import { APIOrganizationCommand, formatAndWriteItem, TableGenerator } from '@smartthings/cli-lib'
 
-import { CapabilityId, capabilityIdOrIndexInputArgs, CapabilitySummaryWithNamespace, getCustomByNamespace, translateToId } from '../../lib/commands/capabilities-util'
+import { capabilityIdOrIndexInputArgs, chooseCapability } from '../../lib/commands/capabilities-util'
 
 
 export function buildTableOutput(tableGenerator: TableGenerator, presentation: CapabilityPresentation): string {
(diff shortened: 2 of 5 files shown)


packages/cli/src/commands/capabilities/presentation.ts
```diff
@@ -82,19 +82,17 @@ export default class PresentationsCommand extends APIOrganizationCommand<typeof
...
-			sortKeyName,
-			listTableFieldDefinitions: ['id', 'version', 'status'],
-			buildTableOutput: (data: CapabilityPresentation) => buildTableOutput(this.tableGenerator, data),
+
+		const capabilityId = await chooseCapability(this, this.args.id, this.args.version, undefined, this.flags.namespace)
+		try {
+			const presentation = await this.client.capabilities.getPresentation(capabilityId.id, capabilityId.version)
+			await formatAndWriteItem(this, { buildTableOutput: data => buildTableOutput(this.tableGenerator, data) }, presentation)
+		} catch (error) {
+			if ('message' in error && error.message.includes('status code 403')) {
```
rossiam, line 91: I usually avoid special error handling of errors from the API. The API returns error messages in a standard format (most of the time). We have an issue open to "display standard error messages in a nicer format" which would do something akin to this, but generically for all commands and errors.

Also, if you do want to check for a status code, you should be able to find it more specifically at `error.response.status` like this:

```if (error.response?.status === 403)```

File: packages/cli/src/commands/deviceprofiles/view.ts [hunk 2]
```diff
@@ -1,8 +1,6 @@
-import { DeviceProfile } from '@smartthings/core-sdk'
+import { APIOrganizationCommand, formatAndWriteItem, outputItemOrList } from '@smartthings/cli-lib'
```
rossiam, line 1: Similar to `presentation.ts`, we shouldn't need `outputItemOrList` here anymore and the flags should come from `formatAndWriteItem`.

rossiam (approved):

.changeset/violet-moles-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@smartthings/cli": patch
+---
+
+fix `capabilities:presentation` and `deviceprofiles:view` commands to wait for user selection
```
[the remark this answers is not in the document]
Sitlintac: That's much better, thanks!

packages/cli/src/commands/capabilities/presentation.ts
```diff
@@ -2,9 +2,9 @@ import { Flags } from '@oclif/core'
 
 import { CapabilityPresentation } from '@smartthings/core-sdk'
 
-import { APIOrganizationCommand, OutputItemOrListConfig, outputItemOrListGeneric, TableGenerator } from '@smartthings/cli-lib'
+import { APIOrganizationCommand, formatAndWriteItem, outputItemOrListGeneric, TableGenerator } from '@smartthings/cli-lib'
```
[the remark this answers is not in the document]
Sitlintac: It didn't even occur to me to switch out the flags when I changed the output method.  I'll remember to watch out for this in future, too.

packages/cli/src/commands/capabilities/presentation.ts
Sitlintac: I've taken out the special handling.  I really prefer having a nicer format to the message, but I see your point that it's better handled uniformly via that other open issue.

packages/cli/src/commands/deviceprofiles/view.ts
Sitlintac: Done.

PR #513, opened 2023-05-11
fix: don't include unenrolled channels and hubs when prompting user for edge:channels:unenroll

rossiam:
* when using prompts for unenrolling a hub from a channel
  * prompt for hub first, skipping hubs that aren't enrolled in any channels
  * only include enrolled channels when prompting for channel
* added examples to `edge:channels:unenroll` since I was updating it
* changes to library to support this
  * `ChooseOptions` type now includes optional `listItems` so `chooseX` methods can easily support alternate lists
  * `ChooseOptions` is now generic so `listItems` can require function that returns list of the correct things
  * specify the generic where `ChooseOptions` is used
  * updated `chooseHubs` and `chooseChannel` to support alternate `listItems`
  * unit tests for updated for `chooseHubs` and `chooseChannel`

## Checklist


packages/cli/README.md
```diff
@@ -3384,6 +3384,20 @@ COMMON FLAGS
...
+EXAMPLES
+  prompt user for hub and channel and then unenroll hub from channel
+
+    $ smartthings edge:channels:unenroll
+
+  prompt user for a channel the given hub is enrolled in and then unenroll it
+
+    $ smartthings edge:channels:unenroll 2fa0912f-cb73-424e-97f9-ffff76ea4f2a
+
+  unenroll the hub from the specified channel
```
Doczillar, line 3397: What is the difference between these two examples? The first seems to be showing unenrolling a hub from a channel, and the second is "unenroll[ing] the hub from the specified channel".

Sitlintac (approved):

packages/cli/README.md
Sitlintac: The first example doesn't provide any arguments, so the user will be shown a list of hubs and prompted to choose one, then shown a list of channels that hub is subscribed to and prompted to choose one.  The second example gives one argument -- the UUID of a hub -- so it'll skip straight to asking for the channel to unsubscribe that specific hub from. The third example specifies the UUID of both the hub and the channel to unenroll it from, so the user won't be prompted for anything else.

packages/edge/src/lib/commands/drivers-util.ts
```diff
@@ -14,6 +14,7 @@ import {
 	chooseOptionsDefaults,
 	chooseOptionsWithDefaults,
 	forAllOrganizations,
+	ListDataFunction,
```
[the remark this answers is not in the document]
rossiam: Oops, fixed.

PR #518, opened 2023-05-25
chore: switch to ES modules

rossiam:
* switched to using ES modules
  * updated imports to include ".js" suffix as required by ES modules
  * updated config for jest to work with ES modules too
* updated eslint and related dependencies
* updated jest and related dependencies
* tests are broken in the cli package but I will deal with that in the next pull request when I make this just a single package project
* shared jest config was copied into sub-projects to get jest to work for now. These will all be merged into a single file in the next pull request anyway.

## Checklist


packages/cli/src/commands/apps/update.ts
```diff
@@ -17,7 +17,7 @@ import {
 	InputDefsByProperty,
 } from '@smartthings/cli-lib'
 import { addPermission } from '../../lib/aws-utils'
-import { chooseApp, smartAppHelpText, tableFieldDefinitions } from '../../lib/commands/apps-util'
+import { chooseApp, tableFieldDefinitions } from '../../lib/commands/apps-util'
```
Sitlintac, line 20: Do these need the `.js` extension?

Sitlintac (approved):

packages/cli/src/commands/apps/update.ts
rossiam: Amazing catch. They will but we can fix them when we update and move them since they aren't being compiled currently. I'm not sure how my regex search and replaces missed them!

PR #525, opened 2023-06-16
refactor: convert config command to yargs format

rossiam:
# Overview

Added `config` command as first command translated to yargs. The `config` was chosen because it has the fewest module dependencies to keep this pull request from being even larger.

NOTE: For now, to run, use `ts-node`. Either install globally via `npm i -g ts-node` and run `ts-node /path/to/cli-repo/dist/run.js` or run `npx ts-node /path/to/cli-repo/dist/run.js`.

In the old oclif world, we have several functions that directly support common types commands like `outputItemOrList` and `inputAndOutputItem`. These functions all have an associated oclif-specific `flags` property which we use in the commands to ensure flags used by the function are available to the function, usually via its `command` argument.

In this new world, these functions now have two parts shared for the same reason. First, there is a `builder` function, nested in its parent function as a property, which replaces the old `flags` property and can contribute to the `yargs` configuration (usually by adding flags). The second is a type named after the function, starting with a capital letter and with the word `Flags` appended. The `yargs` system can determine this from the builder so we could technically leave it out but I found it's really helpful to define the expected results in this simple type. The end result is that if this definition and what is defined by the builder don't match, we get a compile error. Without specifying these extra types, it's easier to make a mistake writing the `builder` function and not notice it.

# Details of Changes Made

* moved `typings` directory (which contains types for some Jest extenstions) into the `__test__` directory and removed old directory from jest `coveragePathIgnorePatterns` (new directory already excluded by `'/__tests__/',` ignore pattern)
* added yargs dependency
* removed cli-config.ts mock file (We don't need this now and we can add it back if we decide we do later.)
* fixed version number in package.json that didn't get updated in rebase
* removed `@log4js-node/log4js-api` and used `@types//log4j` instead; also moved to dev dependencies where it should have been in the first place, updated some code for this
* updated a couple of eslint dependencies
* added src/commands/index.ts to list commands for yargs (currently only contains `config` command)
* moved and converted the following files:
  * smartthings-command
    * translated to yargs format
      * turned `SmartThingsCommand` class into `smartThingsCommand` function
      * renamed `SmartThingsCommandInterface` to `SmartThingsCommand`
      * addded `SmartThingsCommandFlags` and `smartThingsCommandBuilder`
    * removed `cancel` and `exit`, at least for now
  * config command itself (appears as a delete and add)
    * moved and translated to yargs command
    * removed `reservedKey` handling since we don't keep logging info in the main config file anymore
    * converted `ConfgItem` from class to a `type`
    * minor clean-up
    * added examples
  * output-builder.ts
  * basic-io.ts; see "upcoming refactors" below for TODO about standardizing error handling
  * command-util.ts
  * format.ts
    * fixed types a bit
  * input-builder.ts
  * input.ts -> input-processor.ts
    * renamed
	* moved "common IO" flags to `output.ts` and `output-builder.ts` (they are really only "O" flags)
  * listing-io.ts
  * output.ts
* moved `file-util.ts`, `user-query.ts`, `util.ts`, and `validate-util.ts` into `lib` subdirectory
* removed `moduleResolution` from `tsConfig-base.json` which was causing problems running.
* added `run.ts` and updated `index.ts` to configure and run CLI (with only `config` command so far) using `yargs`
* added src/commands/index.ts which will list all commands for `yargs` (`yargs` with ESM doesn't support auto discovery of commands in a directory)
* moved "library"-type modules into 'lib' subdirectory (also moved associated tests)

# Upcoming refactors

  * move `stringConfigValue` and its ilk into `cli-config.ts`, maybe keep proxies
  * move basic types at the top of from `basic-io.ts` to some more general place like `types.ts`
  * standardize error handling once we have a few commands converted
  * consolidate profile stubbing in tests once we have a few commands
  * move `tableGenerator` from `smartthings-command.ts` to `output-builder.ts` or even just to where it's used; remove `groupRows` from `smartThingsCommand`.
  * don't pass whole command around so much (e.g. into `formatAndWriteItem` or `formatAndWriteList`)
  * (maybe) use functions instead of class in `table-generator.ts` for `DefaultTableGenerator`
  * (maybe) use types and functions instead of classes classes in `input.ts`
  * (maybe) move calculating indent out of `output-bulider.ts` into someplace more general
  * `api-helpers.test.ts` doesn't need to use spies
  * consider referring to "common" input and output as "friendly" instead
  * fill in some unit tests; in particular `smartthings-command.test.ts` (after other refactors, in particular `table-gnerator.ts`)

# Upcoming Fixes

  * VSCode doesn't recognize Jest extensions
  * configDir and cacheDir need to be platform-independent


src/__tests__/lib/command/smartthings-command.test.ts
```diff
@@ -0,0 +1,161 @@
...
+describe('smartThingsCommand', () => {
+	const defaultLogConfig = { config: 'default' } as unknown as log4js.Configuration
+	const buildDefaultLog4jsConfigMock = jest.mocked(buildDefaultLog4jsConfig).mockReturnValue(defaultLogConfig)
+	const logConfig = { config: 'final' } as unknown as log4js.Configuration
+	const loadLog4jsConfigMock = jest.mocked(loadLog4jsConfig).mockReturnValue(logConfig)
+	const log4jsConfigureMock = jest.mocked(log4js.configure)
+	const traceMock = jest.fn()
+	const loggerMock = {
+		trace: traceMock,
+	} as unknown as log4js.Logger
```
Sitlintac, line 23: I'm curious why `loggerMock` got formatted as multi-line instead of a single line.  It looks structurally similar to me to the definition for `logConfig`, which is single-line.

src/__tests__/lib/command/smartthings-command.test.ts
```diff
@@ -0,0 +1,161 @@
...
+	const log4jsGetLoggerMock = jest.mocked(log4js.getLogger).mockReturnValue(loggerMock)
+	const cliConfig = { profile: {} } as CLIConfig
+	const loadConfigMock = jest.mocked(loadConfig).mockResolvedValue(cliConfig)
+
+	it('works', async () => {
```
Sitlintac, line 28: LOL at this test name!

Sitlintac (approved):
Fantastic work!  I'm excited to see this branch grow.

src/lib/io-util.ts
```diff
@@ -1,4 +1,4 @@
-import log4js from '@log4js-node/log4js-api'
+import log4js from 'log4js'
 import fs from 'fs'
 import path from 'path'
 import yaml from 'js-yaml'
```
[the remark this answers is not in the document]
rossiam: Those are good words but they could also apply to JSON or YAML output which can also go to the terminal. I'm looking for a word that indicates the more user-friendly version of both input and output (which is why I'm considering "friendly"). Currently this is always table-formatted output and Q&A input but I could see using something other than table output in some situations.

src/__tests__/lib/command/smartthings-command.test.ts
rossiam: When I wrote it I thought I was going to add more than what I ended up with. I've condensed this back to one line to match the format of `logConfig`.

src/__tests__/lib/command/smartthings-command.test.ts
rossiam: 😂  Oops...that was not meant to stay in! I almost always start with `works` for my first test and fix it. I think I'll leave it until I flesh these tests out more though.

PR #532, opened 2023-07-13
feat: add support for schema app invitations

rossiam:
Add support for schema invitations.

## Checklist

### DIFF
--- a/packages/cli/src/commands/invites/schema/check-acceptance.ts
+++ b/packages/cli/src/commands/invites/schema/check-acceptance.ts
@@ -1,0 +1,45 @@
+import { Flags } from '@oclif/core'
+
+import { APICommand, FormatAndWriteItemConfig, formatAndWriteItem } from '@smartthings/cli-lib'
+
+import { chooseSchemaInvitation } from '../../../lib/commands/invites-utils'
+import { SchemaAppAcceptanceStatus } from '@smartthings/core-sdk'
+
+
+export default class DeviceStatusCommand extends APICommand<typeof DeviceStatusCommand.flags> {
+	static description = 'check the acceptance status of a schema app invitation'
+
+	static flags = {
+		...APICommand.flags,
+		...formatAndWriteItem.flags,
+		// eslint-disable-next-line @typescript-eslint/naming-convention
+		'schema-app': Flags.string({
+			description: 'schema app id',
+		}),
+	}
+
+	static args = [{
+		name: 'id',
+		description: 'the invitation id',
+	}]
+
+	static examples = [
+		{
+			description: 'prompt for an invitation and check its acceptance status',
+			command: 'smartthings invites:schema:check-acceptance',
+		},
+		{
+			description: 'check acceptance status for the specified invitation',
+			command: 'smartthings invites:schema:check-acceptance 7bd4c5b6-e840-44b3-9933-549a342d95ce',
+		},
+	]
+
+	async run(): Promise<void> {
+		const id = await chooseSchemaInvitation(this, this.flags['schema-app'], this.args.idOrIndex)
+		const status = await this.client.invitesSchema.getAcceptanceStatus(id)
+		const config: FormatAndWriteItemConfig<SchemaAppAcceptanceStatus> = {
+			tableFieldDefinitions: ['isAccepted', 'description', 'shortCode'],
+		}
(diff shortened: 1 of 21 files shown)


Sitlintac (approved):

packages/cli/src/commands/invites/schema/check-acceptance.ts
```diff
@@ -0,0 +1,45 @@
...
+		description: 'the invitation id or number in list',
+	}]
+
+	static examples = [
+		{
+			description: 'prompt for an invitation and check its acceptance status',
+			command: 'smartthings invites:schema:check-acceptance',
+		},
+		{
+			description: 'check its acceptance status for the specified invitation',
```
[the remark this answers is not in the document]
rossiam: Oops, fixed.

packages/cli/src/commands/invites/schema/delete.ts
```diff
@@ -0,0 +1,44 @@
...
+		{
+			description: 'delete a specific invitation by id',
+			command: 'smartthings invites:schema:delete 5dfd6626-ab1d-42da-bb76-90def3153998',
+		},
+	]
+
+	async run(): Promise<void> {
+		const id = await chooseSchemaInvitation(this, this.flags['schema-app'], this.args.id)
+		await this.client.invitesSchema.revoke(id)
+		this.log(`Invitation ${id} revocation requested. (This may take a few moments to complete.)`)
```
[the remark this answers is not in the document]
rossiam: I've changed the wording to use "deletion requested". The docs use both "revoke" and "delete" so I think we get to pick.

packages/cli/src/commands/invites/schema/create.ts
```diff
@@ -0,0 +1,79 @@
...
+		},
+	]
+
+	async getInputFromUser(): Promise<SchemaAppInvitationCreate> {
+		// Save selected schema apps so we can access their name quickly for summarizeForEdit below.
+		const schemaAppsById = new Map<string, SchemaApp>()
+
+		// TODO: allow default; write issue
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		const updateFromUserInput = async (original?: string): Promise<string | CancelAction> => {
```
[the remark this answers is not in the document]
rossiam: No, I just shouldn't have included it. Removed now.

packages/lib/src/user-query.ts
```diff
@@ -74,34 +74,69 @@ export const askForString = async (message: string, options?: AskForStringOption
...
+	validate?: ValidateFunction
+
+	helpText?: string
+}
+
+const promptForInteger = async (message: string, options: AskForIntegerOptions): Promise<number | undefined> => {
+	const buildValidateFunction = (): ValidateFunction | undefined => {
+		// When there is a validate method defined and we have help text, we have to allow '?'.
+		if (options.helpText && options.validate) {
+			return input => input === '?' || options.validate === undefined || options.validate(input)
```
Sitlintac, line 94: Since this is inside an if statement that requires `options.validate` to be truthy, why does the returned function include `options.validate === undefined`?

packages/lib/src/user-query.ts
```diff
@@ -74,34 +74,69 @@ export const askForString = async (message: string, options?: AskForStringOption
...
+		console.log(options.helpText)
+		entered = await prompt()
+	}
+
+	return entered ? Number(entered) : undefined
+}
+
+/**
+ * Simple wrapper around querying a user for a string. The return value will always be a string
+ * with at least one character or undefined.
```
Sitlintac, line 120: Copy/paste oopsie: we're now querying for an optional integer, not a string. (And the next function is querying for a [mandatory] integer.)

Sitlintac (approved):

packages/lib/src/user-query.ts
rossiam: Good eye but it is necessary to make TypeScript happy because it's actually *not* wrapped in that `if` anymore when it gets called. The `buildValidateFunction` function returns a new one-line function validate function for use later. I violated our rule about naming functions that return functions with `Fn` so I've fixed that though.

packages/lib/src/user-query.ts
rossiam: Another good catch. Fixed.

bflorian (approved):

PR #541, opened 2023-08-23
fix: make cancel action actually cancel even when there are validation errors

rossiam:
* Fixed cancel action to actually cancel even when validation errors exist.

## Checklist

### DIFF
--- a/packages/lib/src/item-input/command-helpers.ts
+++ b/packages/lib/src/item-input/command-helpers.ts
@@ -1,6 +1,7 @@
 import inquirer, { ChoiceCollection } from 'inquirer'
 
 import { jsonFormatter, OutputFormatter, yamlFormatter } from '../output'
+import { red } from '../colors'
 import { SmartThingsCommandInterface } from '../smartthings-command'
 import {
 	cancelAction,
@@ -11,7 +12,6 @@ import {
 	previewJSONAction,
 	previewYAMLAction,
 } from './defs'
-import { red } from 'chalk'
 
 
 export type UpdateFromUserInputOptions = {
@@ -55,9 +55,10 @@ export const updateFromUserInput = async <T extends object>(command: SmartThings
 		if (validationResult !== true) {
 			console.log(red(validationResult))
 			const answer = await inputDefinition.updateFromUserInput(retVal)
-			if (answer !== cancelAction) {
-				retVal = answer
+			if (answer === cancelAction) {
+				command.cancel()
 			}
+			retVal = answer
 			continue
 		}
 		const choices: ChoiceCollection = [
@@ -76,7 +77,7 @@ export const updateFromUserInput = async <T extends object>(command: SmartThings
 			name: 'action',
 			message: 'Choose an action.',
 			choices,
-			default: validationResult === true ? finishAction : editAction,
+			default: finishAction,
 		})).action
 
 		if (action === editAction) {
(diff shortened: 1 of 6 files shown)


packages/lib/src/item-input/command-helpers.ts
```diff
@@ -11,7 +12,6 @@ import {
 	previewJSONAction,
 	previewYAMLAction,
 } from './defs'
-import { red } from 'chalk'
```
rossiam, line 14: Using our `colors` proxy module makes unit testing easier.

packages/lib/src/item-input/command-helpers.ts
```diff
@@ -76,7 +77,7 @@ export const updateFromUserInput = async <T extends object>(command: SmartThings
 			name: 'action',
 			message: 'Choose an action.',
 			choices,
-			default: validationResult === true ? finishAction : editAction,
```
rossiam, line 79: When working on unit tests (still in progress), I noticed there is no way to get to this point if `validationResult` is not true.

packages/lib/src/item-input/object.ts
```diff
@@ -190,10 +190,6 @@ export function objectDef<T extends object>(name: string, inputDefsByProperty: I
 							} else if (propertyName === updatedPropertyName) {
 								afterUpdatedProperty = true
 							}
-							const propertyInputDefinition = inputDefsByProperty[propertyName]
-							if (!propertyInputDefinition) {
-								continue
-							}
```
rossiam, line 196: I noticed this code wasn't doing anything while I was working on unit tests.

packages/lib/src/item-input/object.ts
```diff
@@ -228,10 +224,6 @@ export function objectDef<T extends object>(name: string, inputDefsByProperty: I
 								// (Once this is implemented, remove note from documentation for this function.)
 								afterUpdatedProperty = true
 							}
-							const propertyInputDefinition = inputDefsByProperty[propertyName]
-							if (!propertyInputDefinition) {
-								continue
-							}
```
rossiam, line 234: I noticed this code wasn't doing anything while I was working on unit tests.

packages/lib/src/item-input/object.ts
```diff
@@ -218,7 +214,7 @@ export function objectDef<T extends object>(name: string, inputDefsByProperty: I
 								const laterPropertyInputDefinition = inputDefsByProperty[propertyName]
 								const updateIfNeeded = laterPropertyInputDefinition.updateIfNeeded
 								if (updateIfNeeded) {
-									const laterPropertyValue = await updateIfNeeded(updated[propertyName], updatedPropertyName, [{ ...updated }, ...context])
+									const laterPropertyValue = await updateIfNeeded(updated[propertyName], action, [{ ...updated }, ...context])
```
rossiam, line 217: Another thing I noticed while working on the unit tests. The `updatedPropertyName` is fine for top-level properties (line 105) but for nested properties, it only tells us which object the nested property is in and not property name inside that object that is actually being edited.

A real-world example of this is the `clientName` definition on line 58 of the `apps:create` command (`packages/cli/src/commands/apps/create.ts`). Before this fix, it would pass just `oauth` but after this it will pass `oauth.clientName`.

(There is currently no code that uses this so the difference doesn't matter ATM but we should be passing on everything.)

Sitlintac (approved):

PR #566, opened 2023-11-29
fix: minor history fixes

rossiam:
* treat integer values for `before` and `after` flags as seconds since the epoch as describe in the help
* refactor: simplified sort function
* refactor: minor reformatting
* refactor: use `confirm` inquirer input type for y/n question instead of text input
* in JSON/YAML output, strictly obey `after`
  * the API first returns items after the specified `after` but allows the user to continue on with the "next" page link
  * the interactive table format output still allows the user to continue on to view items, as the API does
  * when asking for JSON or YAML output, we now strictly obey the `after` flag

## Checklist

### DIFF
--- a/packages/cli/src/lib/commands/history-util.ts
+++ b/packages/cli/src/lib/commands/history-util.ts
@@ -36,17 +36,15 @@ export const historyFlags = {
 	}),
 }
 
-export function toEpochTime(date?: string): number | undefined {
-	if (date) {
-		return new Date(date).getTime()
-	}
-}
+export const toEpochTime = (date?: string): number | undefined =>
+	date ? (date.match(/^\d+$/) ? Number(date) : new Date(date).getTime()) : undefined
 
-export function sortEvents(list: DeviceActivity[]): DeviceActivity[] {
-	return list.sort((a, b) => a.epoch === b.epoch ? 0 : (a.epoch < b.epoch ? 1 : -1))
-}
+export const sortEvents = (list: DeviceActivity[]): DeviceActivity[] => list.sort((a, b) => b.epoch - a.epoch)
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/commands/history-util.ts
```diff
@@ -64,36 +62,37 @@ export function getNextDeviceEvents(table: Table, items: DeviceActivity[], optio
...
-				}
+				message: 'Fetch more history records?',
+				default: true,
+			})).more as boolean
+
+			if (!more) {
+				break
+			}
+
+			const table = command.tableGenerator.newOutputTable({ isList: true, head })
```
Sitlintac, line 91: I didn't know you could re-define a `const` in a more restricted context like this.  Neat.  But could certainly get confusing in a less straightforward situation than this one.

packages/cli/src/lib/commands/history-util.ts
```diff
@@ -132,6 +131,11 @@ export const getHistory = async (client: SmartThingsClient, limit: number, perRe
 	const items: DeviceActivity[] = [...history.items]
 	while (items.length < limit && history.hasNext()) {
 		await history.next()
+		// The API allows the user to continue to records before the specified after with paging so
```
Sitlintac, line 134: This wording is a little hard to follow. I think it's "continue to records before the specified after" that confuses me--maybe something like "continue to retrieve records before the specified 'after' date"?

Sitlintac (approved):

packages/cli/src/lib/commands/history-util.ts
rossiam: Confusing? That's an understatement. Mangled is more like it. 😄 I must have messed this in editing. I've updated it to:

```
		// The API allows the user to continue to view history from before the specified "after"
		// with paging so we stop processing if we get items that come before the specified "after".
```

PR #579, opened 2024-01-05
docs: update help text for "accept limit" for schema app creation

rossiam:
Updated help text for the `acceptLimit` field in schema invitation creation to match new limit and new information.


packages/cli/src/commands/invites/schema/create.ts
```diff
@@ -75,10 +75,12 @@ export default class InvitesSchemaCreateCommand extends APICommand<typeof Invite
...
 			description: optionalStringDef('Description'),
-			acceptLimit: optionalIntegerDef('Accept Limit',
+			acceptLimit: optionalIntegerDef('Accept (Install) Limit',
 				{
-					helpText: 'Enter the maximum number of users you want to be able to accept the invitation.\n' +
-						'Leave blank for no maximum.',
+					helpText: 'Enter the maximum number of installs you want to allow for the invitation.\n' +
+						'This limits the number of installs but we use the "accept" terminology here to match the API call.\n' +
+						'There is also a limit of 30 days that cannot be changed. After that you will need to create a new invitation.\n' +
+						'Leave blank for a maximum of 500.',
```
Sitlintac, line 83: Here's a suggestion for alternate wording:
> Invitations are limited to a maximum of 500 installations (called "accept"s by the API) and are only valid for 30 days.
> Once either the install limit or the time limit is reached, you will need to create a new invitation.
> Leave blank to accept the maximum install limit or enter a smaller cap if desired.
The format of this suggested wording may be too different from our other help text to be used.  And I'm not entirely sure it's better than what you already have.  But I think having so much other information between "Enter the..." and "Leave blank for..." has the potential to be confusing.

Sitlintac (approved):

packages/cli/src/commands/invites/schema/create.ts
rossiam: I double-checked and 500 is just the default, not the absolute maximum. How about this text?

> Invitations are limited to a maximum of 500 installations by default (called "accept"s by the API) and are only valid for 30 days.
> Once either the install limit or the time limit is reached, you will need to create a new invitation.
> Leave blank to accept the default or enter a different install limit if desired.

Doczillar: If the API doesn't mention the word "installations" anywhere, would it be beneficial to forgo using "installations" in the CLI help text to avoid introducing a new term? 
> Invitations are limited to a maximum of 500 acceptances by default and are valid for 30 days. 
Once either the accept limit or the time limit is reached, you will need to create a new invitation.
Leave blank to accept the default limit of 500, or enter a different acceptance limit if desired.

I do think the word "installations" is more accurate here; I'm thinking a user could accept an invitation and install an app multiple times.

Sitlintac: Maybe add an extra sentence that defines "acceptance" so it's clear it's per-installation and not per-user?

> An acceptance is when a user installs an app from your invitation; multiple installations count as multiple acceptances.
> Invitations are limited to a maximum of 500 acceptances by default and are valid for 30 days.
> Once either the accept limit or the time limit is reached, you will need to create a new invitation.
> Leave blank to accept the default limit of 500, or enter a different acceptance limit if desired.

Doczillar (approved):

PR #584, opened 2024-01-17
test: convert listing-io, output-builder, and select tests to ESM

rossiam:
converted listing-io, output-builder, and select tests to ESM

### DIFF
--- a/src/__tests__/lib/command/output-builder.test.ts
+++ b/src/__tests__/lib/command/output-builder.test.ts
@@ -1,41 +1,59 @@
-import yargs, { Argv } from 'yargs'
+import { jest } from '@jest/globals'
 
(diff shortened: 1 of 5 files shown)


src/__tests__/lib/command/output-builder.test.ts
```diff
@@ -1,41 +1,59 @@
...
+	calculateOutputFormatBuilder: calculateOutputFormatBuilderMock,
+	jsonFormatter: jsonFormatterMock,
+	yamlFormatter: yamlFormatterMock,
+}))
+
+
+const { buildOutputFormatter, buildOutputFormatterBuilder } = await import('../../../lib/command/output-builder.js')
+
 
 test('buildOutputFormatterBuilder', () => {
```
rossiam, line 34: This is the only unit test I've written so far for a "builder" function. I'm not sure if it's really worth it but I might try a couple more and see how they look. If nothing else, it's nice to get that sweet sweet 100% coverage. 😄

Sitlintac (approved):

PR #593, opened 2024-02-26
refactor: convert locations command to yargs

rossiam:
* fixed typo in jest config
* moved files for `locations` command out of old `packages` into `src`
* created new `createChooseFn` function which creates a `create<Item>`-type function and replaced code for `chooseApp` and `chooseLocation` with a call to this function
* moved utility functions out of `locations.ts` into `locations-util.ts`
* converted `locations` command to yargs
* added or converted unit tests for everything

Related work coming up soon (ordering is intentional):

1. finish unit tests for `apps` and `devicepreferences` commands
2. refactor mocks for builders into a test utility like we did for logger

### DIFF
--- a/src/__tests__/commands/locations.test.ts
+++ b/src/__tests__/commands/locations.test.ts
@@ -1,75 +1,135 @@
-import { outputItemOrList, selectFromList } from '@smartthings/cli-lib'
-import LocationsCommand, { chooseLocation } from '../../commands/locations.js'
-import { LocationsEndpoint } from '@smartthings/core-sdk'
-import { Config } from '@oclif/core'
+import { jest } from '@jest/globals'
+
+import { ArgumentsCamelCase, Argv } from 'yargs'
+
+import { Location, LocationsEndpoint, SmartThingsClient } from '@smartthings/core-sdk'
+
+import { APICommand, APICommandFlags, apiCommand, apiCommandBuilder, apiDocsURL } from '../../lib/command/api-command.js'
+import { outputItemOrList, outputItemOrListBuilder } from '../../lib/command/listing-io.js'
+import { CommandArgs } from '../../commands/locations.js'
 
 
-const listSpy = jest.spyOn(LocationsEndpoint.prototype, 'list').mockImplementation()
 
-describe('chooseLocation', () => {
-	const mockSelectFromList = jest.mocked(selectFromList)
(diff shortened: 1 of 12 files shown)


src/__tests__/commands/locations.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+	const builder = cmd.builder as (yargs: Argv<object>) => Argv<CommandArgs>
+	expect(builder(yargsMock)).toBe(outputItemOrListArgvMock)
+
+	expect(positionalMock).toHaveBeenCalledTimes(1)
+	expect(exampleMock).toHaveBeenCalledTimes(1)
+	expect(epilogMock).toHaveBeenCalledTimes(1)
+})
+
+describe('handler', () => {
```
Sitlintac, line 65: Here and elsewhere: How are we distinguishing between `test()` and `describe()`, if at all?

Sitlintac (approved):

src/lib/command/util/util-util.ts
```diff
@@ -0,0 +1,34 @@
...
+	config: SelectFromListConfig<T>,
+	listItems: (client: SmartThingsClient) => Promise<T[]>,
+): ChooseFunction<T> =>
+	async (
+			command: APICommand<SelectFromListFlags>,
+			locationFromArg?: string,
+			options?: Partial<ChooseOptions<T>>): Promise<string> => {
+		const opts = chooseOptionsWithDefaults(options)
+
+		const listItems2 = (): Promise<T[]> => listItems(command.client)
```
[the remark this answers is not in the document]
rossiam: Oh, shoot. I didn't mean to leave that. I gave it that name because I didn't want to break my stride and I had planned to go back and fix it. It's fixed now.

BTW, I like `Wrapper` too. I don't think I would have come up with that on my own. I probably would have done something like named the argument `listItemsIn` and given this variable `listItems` but I like the `Wrapper` suffix better so I'm kinda glad I forgot. 😄

src/__tests__/commands/locations.test.ts
rossiam: `describe` wraps one or more tests and `test` (or its alias `it`), _is_ a test. What I have been doing is use `test('functionName, ...` when there is only one test to write for a function and `describe` with multiple `it('does this thing', ...` when a function needs more than one test.

PR #619, opened 2024-09-10
feat: query user for organization in schema:update and schema:create commands

rossiam:
* Updated core SDK version that supports including organization id as a parameter to the schema update command.
* Added new `selectDef` definition type that allows the user to select a single item from a list. I plan to write unit tests for this on the yargs branch rather than writing them here and then translating them.
* Restricted `finishVerb` to just "create" and "update" in command helpers because we also need a noun version in one place.
* `schema-utils.ts`
  * Added `organizationDef` function to create the input definition for the organization and called it from 
  * `buildInputDefinition` is now async so it can look up organizations.
  * `buildInputDefinition` now includes the chosen organization id in the return.
  * Unit tests for this were updated enough to get them to work but I plan to do better on the yargs branch.
* `schema:create` and `schema:update` commands were updated to pass on the chosen organization to the core SDK. Unit tests for these were update enough to work but I plan to do better on the yargs branch.


packages/cli/src/commands/schema/update.ts
```diff
@@ -44,10 +51,13 @@ export default class SchemaUpdateCommand extends APIOrganizationCommand<typeof S
 
 		const getInputFromUser = async (): Promise<SchemaAppRequest> => {
 			const original = await this.client.schema.get(id)
+			if (original.certificationStatus === 'wwst' || original.certificationStatus === 'cst') {
+				this.cancel('Schema apps that have already been certified cannot be updated.')
```
Sitlintac, line 55: Consider "...cannot be updated via the CLI."

packages/lib/src/item-input/select.ts
```diff
@@ -0,0 +1,83 @@
...
+	helpText?: string
+}
+
+export type SelectDefChoice<T> =
+	| string
+	| number
+	| { name: string; value: T }
+
+/**
+ * InputDefinition choosing a single item from a list.
```
Sitlintac, line 27: I think it would scan better as "InputDefinition for choosing a single item from a list."

Sitlintac (approved):

packages/cli/src/lib/commands/schema-util.ts
```diff
@@ -73,7 +79,29 @@ export const validateFinal = (schemaAppRequest: InputData): true | string => {
...
+		return staticDef(organizations[0].organizationId)
+	}
+
+	const choices = organizations
+		.map(organization => ({
+			name: organization.name,
+			value: organization.organizationId,
+		}))
+
+	const helpText = 'The organization to which the Schema connector should be associated.'
```
Doczillar, line 96: The smallest of nits:
```suggestion
	const helpText = 'The organization with which the Schema connector should be associated.'
```
I am not 100% confident on this one, but I'm wondering if this should be "with" instead of "to" - @Sitlintac what do you think?

Sitlintac: I considered that question too and left it as-is because I couldn't decide if this was one of those instances where "associated to" is correct.  I think it might be (if you consider an organization to be of higher rank than a connector, which is plausible to me because each connector belongs to one organization but an organization can have multiple connectors), but I agree that "with" is almost always the better choice.

packages/cli/src/commands/schema/update.ts
rossiam: Fixed

packages/lib/src/item-input/select.ts
rossiam: Yeah, that was bad. Fixed.

packages/cli/src/lib/commands/schema-util.ts
rossiam: I like "with" better. I think I only had "to" in there because I had started out using the verb "connected" instead of "associated". Fixed.

PR #625, opened 2024-09-20
chore: update dependencies and node version

rossiam:
* updated Node to version 22
* updated dependencies
    * skipped inquirer because it will require more work. There is a new version we should switch to soon anyway.
    * eslint held to 8.x version (for now)
        * `eslint-plugin-import` does not support eslint 9 and [probably won't for a while](https://github.com/import-js/eslint-plugin-import/issues/2948#issuecomment-1872563600)
        * we may want to consider alternatives
        * maybe this will help?: https://prog.world/setting-up-eslint-plugin-import-for-new-eslint-version-9/
    * changes to login-authenticator and its tests were ported from main branch because of core SDK update (I will skip merging the change that does this since it's done here)


Sitlintac:
The build failed on ubuntu because many definitions for es-lint rules couldn't be found.  It succeeded on the other platforms.     I skimmed the release notes and the issues for the eslint plugin and nothing jumped out at me; should we just re-run that test to see if it was a fluke?

Sitlintac (approved):

PR #626, opened 2024-09-24
refactor: consolidate all choose function stuff in util-utils.ts

rossiam:
Moved all the supporting types and variables for `createChooseFn` into the same file as `createChooseFn`. Not sure why I didn't do this when I created `createChooseFn`. Also updated affected unit tests.

### DIFF
--- a/src/__tests__/lib/command/util/apps-util.test.ts
+++ b/src/__tests__/lib/command/util/apps-util.test.ts
@@ -2,33 +2,36 @@ import { jest } from '@jest/globals'
 
 import { AppResponse, AppsEndpoint, AppType, PagedApp, SmartThingsClient } from '@smartthings/core-sdk'
 
-import { PropertyTableFieldDefinition, Table, TableGenerator, ValueTableFieldDefinition } from '../../../../lib/table-generator.js'
-import { APICommand } from '../../../../lib/command/api-command.js'
 import {
-	chooseOptionsDefaults,
+	PropertyTableFieldDefinition,
+	Table,
+	TableGenerator,
+	ValueTableFieldDefinition,
+} from '../../../../lib/table-generator.js'
+import { stringTranslateToId } from '../../../../lib/command/command-util.js'
+import {
+	createChooseFn,
+	type ChooseFunction,
 	chooseOptionsWithDefaults,
-	stringTranslateToId,
-} from '../../../../lib/command/command-util.js'
-import { selectFromList, SelectFromListFlags } from '../../../../lib/command/select.js'
+} from '../../../../lib/command/util/util-util.js'
 
 
-const chooseOptionsWithDefaultsMock = jest.fn<typeof chooseOptionsWithDefaults>()
 const stringTranslateToIdMock = jest.fn<typeof stringTranslateToId>()
 jest.unstable_mockModule('../../../../lib/command/command-util.js', () => ({
-	chooseOptionsDefaults,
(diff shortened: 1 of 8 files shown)


src/__tests__/lib/command/util/apps-util.test.ts
```diff
@@ -2,33 +2,36 @@ import { jest } from '@jest/globals'
...
-	chooseOptionsDefaults,
+	PropertyTableFieldDefinition,
+	Table,
+	TableGenerator,
+	ValueTableFieldDefinition,
+} from '../../../../lib/table-generator.js'
+import { stringTranslateToId } from '../../../../lib/command/command-util.js'
+import {
+	createChooseFn,
+	type ChooseFunction,
```
rossiam, line 14: I've recently learned about this `type` marker (even though it's been around a while!) and have started using it in the PR but I'm realizing now that I've done so rather inconsistently. I plan to do better in the next PR and then open up a discussion with the team in standup to talk about it.

src/__tests__/lib/command/util/apps-util.test.ts
```diff
@@ -183,10 +189,8 @@ describe('buildTableOutput', () => {
 		expect(newOutputTableMock).toHaveBeenCalledWith(
 			expect.objectContaining({ head: ['Key', 'Value'] }),
 		)
-		expect(pushMock).toHaveBeenCalledTimes(1)
-		expect(pushMock).toHaveBeenCalledWith(['setting', 'setting value'])
-		expect(toStringMock).toHaveBeenCalledTimes(1)
-		expect(toStringMock).toHaveBeenCalledWith()
+		expect(pushMock).toHaveBeenCalledExactlyOnceWith(['setting', 'setting value'])
+		expect(toStringMock).toHaveBeenCalledExactlyOnceWith()
```
Sitlintac, line 193: I checked the Jest documentation but I'm still not quite sure about something.  Is this in fact checking that `toStringMock` was called exactly once with no parameters?  Or is it not checking parameters since none were supplied?

Sitlintac (approved):

src/__tests__/lib/command/util/apps-util.test.ts
rossiam: A call to a `With` function that doesn't check the `With` seems rather silly so checking for no arguments is the only thing that makes sense to me. It's easy enough to test though; I just removed the argument from a call that had one and I get an error. Here's the output when I remove the argument from line 192 above:

```
 FAIL  src/__tests__/lib/command/util/apps-util.test.ts
  ● buildTableOutput › creates new table with correct options and adds settings

    expect(received).toHaveBeenCalledExactlyOnceWith(expected)

    Expected mock function to have been called exactly once with [], but it was called with ["setting", "setting value"]

      190 | 			expect.objectContaining({ head: ['Key', 'Value'] }),
      191 | 		)
    > 192 | 		expect(pushMock).toHaveBeenCalledExactlyOnceWith()
          | 		                 ^
      193 | 		expect(toStringMock).toHaveBeenCalledExactlyOnceWith()
      194 | 	})
      195 | })

      at Object.<anonymous> (src/__tests__/lib/command/util/apps-util.test.ts:192:20)

Test Suites: 1 failed, 2 passed, 3 total
Tests:       1 failed, 58 passed, 59 total
Snapshots:   0 total
Time:        2.992 s, estimated 6 s
Ran all test suites related to changed files.
```

PR #627, opened 2024-09-26
refactor: move devices:preferences command to yargs

rossiam:
* convert `devices:preferences` command to yargs
* convert `chooseDevice` utility function which is used by `devices:preferences`
    * note that there is more in the old `devices-util.ts` yet to convert that is not yet used by converted commands
    * added support for a filter to `createChooseFn` which is needed by `chooseDevice`
* updated or added unit tests for everything
* in touched unit tests:
    * made use of `toHaveBeenCalledExactlyOnceWith` and `toHaveBeenCalledOnce` from jest.extended
    * converted `.toHaveBeenCalledTimes(0)` to `.not.toHaveBeenCalled()`


julieganzer (approved):
Looks good to me.  Admittedly, getting up to speed.  but i could follow along and seems like you got tests covered!

src/__tests__/commands/devices/preferences.test.ts
```diff
@@ -0,0 +1,154 @@
...
+		},
+		key1: {
+			preferenceType: 'string',
+			value: 'preference string 1',
+		},
+	},
+}
+
+describe('buildTableOutput', () => {
+	it('works', () => {
```
Sitlintac, line 93: Is this test left over from an initial stub or is it still useful?

Sitlintac (approved):

src/__tests__/commands/devices/preferences.test.ts
rossiam: Oops, good catch. I'll remove this in the next pull request.

PR #637, opened 2024-10-24
feat: Force user to choose organization for schema app before doing anything that will automatically assign one

rossiam:
Force user to choose organization for schema app before doing anything that will automatically assign one.

I will update unit tests more after merge to yargs branch.

### DIFF
--- a/packages/cli/src/commands/schema/update.ts
+++ b/packages/cli/src/commands/schema/update.ts
@@ -12,7 +12,11 @@ import {
 } from '@smartthings/cli-lib'
 
 import { addSchemaPermission } from '../../lib/aws-utils'
-import { getSchemaAppUpdateFromUser, SchemaAppWithOrganization } from '../../lib/commands/schema-util'
+import {
+	getSchemaAppEnsuringOrganization,
+	getSchemaAppUpdateFromUser,
+	SchemaAppWithOrganization,
+} from '../../lib/commands/schema-util'
 
 
 export default class SchemaUpdateCommand extends APIOrganizationCommand<typeof SchemaUpdateCommand.flags> {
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/schema/update.ts
```diff
@@ -49,11 +53,12 @@ export default class SchemaUpdateCommand extends APIOrganizationCommand<typeof S
 			listItems: () => this.client.schema.list(),
 		})
 
+		const original = await getSchemaAppEnsuringOrganization(this, id, this.flags)
+		if (original.certificationStatus === 'wwst' || original.certificationStatus === 'cst') {
+			this.cancel('Schema apps that have already been certified cannot be updated via the CLI.')
+		}
+
```
rossiam, line 60: It turns out using the `GET` call for a single schema app isn't the only thing that automatically assigns the user organization by default. It can also happen if you `POST` new JSON so I moved this check out of `getInputFromUser` so we always make sure to have an organization before `POST`ing data from stdin or a file.

dstanich: If the user has a schema app without an org and they go through this path they will encounter this flow?
1. User prompted to pick an org since the schema app doesn't have one
2. CLI will do the GET and associate the schema app to the org
3. This chunk of code will be hit and the user will see an error saying it can't be updated

Is that correct?  [The docs](https://developer.smartthings.com/docs/devices/cloud-connected/publish#associate-an-existing-cloud-connector-with-an-organization) explain to use the `schema:update` command to associate the org ID which it does but then they'd see this error too?

Or if the user doesn't specify anything to update then they won't see it?

rossiam: I've replaced all instances where the `GET` is called to go through this function <sup>1</sup> which most of the time will just ask the user for an organization and then associate it with the app via lines 236-237.

The exception is when we can't ask the user because they are not in a "Q&A" mode. (e.g. They are piping in JSON, reading it from a file, or piping out JSON, etc.) In those cases, we instead log an error which includes a command they can run to fix it.

TLDR, they will either be prompted (when possible) or receive that error; not both.

<sup>1</sup>There is one exception to this. When the user queries an invitation by id, we can still look up the schema app directly without concern because an invitation can't be created for a schema app that doesn't have an organization.

dstanich: Thanks for the adjusted text!  Marking this as resolved.

dstanich (approved):

PR #642, opened 2024-10-29
refactor: add canned mock of exports from api-command

rossiam:
This an example of an attempt at making the unit tests a little less repetitive.

### DIFF
--- a/src/__tests__/commands/apps.test.ts
+++ b/src/__tests__/commands/apps.test.ts
@@ -4,25 +4,18 @@ import { ArgumentsCamelCase, Argv, Options } from 'yargs'
 
 import { AppClassification, AppResponse, AppType, AppsEndpoint, PagedApp, SmartThingsClient } from '@smartthings/core-sdk'
 
-import { APICommand, APICommandFlags, apiCommand, apiCommandBuilder, apiDocsURL } from '../../lib/command/api-command.js'
+import { APICommand, APICommandFlags } from '../../lib/command/api-command.js'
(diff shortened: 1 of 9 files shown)


src/__tests__/commands/apps.test.ts
```diff
@@ -4,25 +4,18 @@ import { ArgumentsCamelCase, Argv, Options } from 'yargs'
...
-
-const apiCommandMock = jest.fn<typeof apiCommand>()
-const apiCommandBuilderMock = jest.fn<typeof apiCommandBuilder>()
-const apiDocsURLMock = jest.fn<typeof apiDocsURL>()
-jest.unstable_mockModule('../../lib/command/api-command.js', () => ({
-	apiCommand: apiCommandMock,
-	apiCommandBuilder: apiCommandBuilderMock,
-	apiDocsURL: apiDocsURLMock,
-}))
+const { apiCommandMock, apiCommandBuilderMock } = apiCommandMocks('../..')
```
rossiam, line 18: I don't love that we need to pass in the directory prefix here but `jest.unstable_mockModules` works relative to the test itself. There might be a way to determine this in the module but I haven't found anything yet.

Sitlintac (approved):

src/__tests__/commands/apps.test.ts
rossiam: Thinking about this for a nanosecond longer and I'm realizing we probably do have more commands at one level deep overall so maybe we should make a default of `../../..`? 🤷

PR #645, opened 2024-11-18
refactor: convert logout command to yargs

rossiam:
Converted `logout` command to `yargs` and added unit tests for it.


src/__tests__/commands/logout.test.ts
```diff
@@ -0,0 +1,92 @@
+import { jest } from '@jest/globals'
+
+import type { ArgumentsCamelCase, Argv } from 'yargs'
+
+import type { Authenticator } from '@smartthings/core-sdk'
+
+import type { APICommand, APICommandFlags } from '../../lib/command/api-command.js'
+import type { CommandArgs } from '../../commands/locations.js'
+import type { fatalError } from '../../lib/util.js'
```
Sitlintac, line 9: Not a type?

Sitlintac (approved):

src/__tests__/commands/logout.test.ts
rossiam: This is only being imported for its type, so `type` is appropriate. You will see this a lot in tests.

PR #650, opened 2024-12-04
refactor: convert apps:oauth:generate command to yargs

rossiam:
* converted `apps:oauth:generate` command to yargs
* added validation requiring at least one scope for an app
* updated and wrote unit tests


src/commands/apps/oauth/generate.ts
```diff
@@ -0,0 +1,87 @@
...
+		{ path: 'oauthClientDetails.scope' },
+		{ path: 'oauthClientDetails.redirectUris' },
+		'oauthClientId',
+		'oauthClientSecret',
+	]
+	const config: InputAndOutputItemConfig<GenerateAppOAuthResponse> = {
+		tableFieldDefinitions,
+	}
+	const getInputFromUser = async (): Promise<GenerateAppOAuthRequest> => {
+		const originalOauth = await command.client.apps.getOauth(appId)
```
Sitlintac, line 65: There's some inconsistency here in how to capitalize OAuth in a variable name: `originalOauth` vs `GenerateAppOAuthRequest`.

src/__tests__/lib/command/util/apps-util-input-primitives.test.ts
```diff
@@ -20,27 +20,31 @@ jest.unstable_mockModule('../../../../lib/item-input/index.js', () => ({
...
+const oauthAppScopeDefMock = buildInputDefMock<string[]>('Scopes Mock')
+checkboxDefMock.mockReturnValueOnce(oauthAppScopeDefMock)
+
+const redirectUriDefMock = buildInputDefMock<string>('Redirect URI Mock')
+stringDefMock.mockReturnValueOnce(redirectUriDefMock)
+const redirectUrisDefMock = buildInputDefMock<string[]>('Redirect URIs Mock')
+arrayDefMock.mockReturnValueOnce(redirectUrisDefMock)
+
+const {
+	oauthAppScopeDef,
```
Sitlintac, line 32: Similar to other comments, there's an alternate option here of `oAuthScopeDef`.  I think I like what you've got better than capitalizing the `A`, but wanted to point it out just in case.

src/__tests__/commands/apps/oauth/generate.test.ts
```diff
@@ -0,0 +1,220 @@
...
+	expect(apiCommandBuilderMock).toHaveBeenCalledExactlyOnceWith(yargsMock)
+	expect(inputAndOutputItemBuilderMock).toHaveBeenCalledExactlyOnceWith(apiCommandBuilderArgvMock)
+
+	expect(positionalMock).toHaveBeenCalledTimes(1)
+	expect(exampleMock).toHaveBeenCalledTimes(1)
+	expect(epilogMock).toHaveBeenCalledTimes(1)
+})
+
+describe('handler', () => {
+	const apiAppsGetOauthMock = jest.fn<typeof AppsEndpoint.prototype.getOauth>()
```
Sitlintac, line 94: Here and elsewhere, there's some inconsistency in how "OAuth" is capitalized when it appears in variable/object/type names.  Here it's treated as one word (`Oauth`) while in line 123, for example, it's treated as two (`OAuth`).

Sitlintac (approved):

src/commands/apps/oauth/generate.ts
rossiam: This is unfortunately inherited from the core SDK which is inconsistent. `GenerateAppOAuthRequest` comes directly from the core SDK and so does `getOauth` which `originalOauth` is mimicking here.

PR #657, opened 2024-12-16
refactor: standarzie on fatalError for handling expected fatal errors

rossiam:
* standardize on using `fatalError` for handling fatal error messages the user can be expected to encounter
* updated unit tests (in some cases I modernized them a bit as well but I stopped doing this at some point)
* some bad unit tests were cleaned up a bit in the process (the tests for `input-and-output-item.ts` were particularly messed up
* changed mocking of `fatalError` to return the string `never return` instead of throwing an exception

### DIFF
--- a/src/__tests__/lib/command/util/capabilities-util.test.ts
+++ b/src/__tests__/lib/command/util/capabilities-util.test.ts
@@ -7,14 +7,22 @@ import type {
 	SmartThingsClient,
 } from '@smartthings/core-sdk'
 
-import type { asTextBulletedList } from '../../../../lib/util.js'
-import { ListDataFunction } from '../../../../lib/command/io-defs.js'
-import { CapabilitySummaryWithNamespace } from '../../../../lib/command/util/capabilities-util.js'
+import type { asTextBulletedList, fatalError } from '../../../../lib/util.js'
+import type { ListDataFunction } from '../../../../lib/command/io-defs.js'
+import { sort } from '../../../../lib/command/output.js'
+import type { CapabilitySummaryWithNamespace } from '../../../../lib/command/util/capabilities-util.js'
 
 
 const asTextBulletedListMock = jest.fn<typeof asTextBulletedList>()
+const fatalErrorMock = jest.fn<typeof fatalError>()
 jest.unstable_mockModule('../../../../lib/util.js', () => ({
 	asTextBulletedList: asTextBulletedListMock,
+	fatalError: fatalErrorMock,
+}))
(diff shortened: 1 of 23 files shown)


src/__tests__/lib/command/util/capabilities-util.test.ts
```diff
@@ -165,32 +174,44 @@ describe('getStandard', () => {
...
-		expect(await translateToId('id', '2', listFunctionMock)).toStrictEqual({ id: 'button', version: 1 })
+		expect(await translateToId('id', '5', listFunctionMock)).toStrictEqual({ id: 'button', version: 1 })
 
-		expect(listFunctionMock).toHaveBeenCalledTimes(1)
-		expect(listFunctionMock).toHaveBeenCalledWith()
+		expect(listFunctionMock).toHaveBeenCalledExactlyOnceWith()
+		expect(sortMock).toHaveBeenCalledExactlyOnceWith(allCapabilitiesWithNamespaces, 'id')
 	})
 
 	it('throws error is out-of-range integer', async () => {
```
Sitlintac, line 204: This existed before this PR, but the description here could be corrected to "throws error when input is out-of-range integer" or "throws error if out-of-range integer".

src/lib/item-input/object.ts
```diff
@@ -43,7 +43,9 @@ export type ObjectDefOptions<T> = {
 	validateFinal?: (item: T, context?: unknown[]) => true | string
 }
 
-const defaultSummarizeForEditFn = (name: string) => (): string => { throw Error(`missing implementation of summarizeForEdit for objectDef ${name}`) }
+const defaultSummarizeForEditFn = (name: string) => (): string => {
+	throw Error(`missing implementation of summarizeForEdit for objectDef ${name}`)
```
Sitlintac, line 47: Just to double check: this does not want to be a `fatalError`?

Sitlintac (approved):

src/__tests__/lib/command/util/capabilities-util.test.ts
rossiam: Good catch. I reworded this a bit to reflect that it's using `fatalError` now too and made the text `fails with fatal error for out-of-range integer`.

src/lib/item-input/object.ts
rossiam: Correct. I intentionally left this one as an exception because, if it were to happen, it would indicates a bug in the code rather than anything the user should see.

PR #675, opened 2025-01-29
refactor: use env-paths library to determine file locations

rossiam:
* made use of `env-path` node library to determine where to locate:
    * config files (files power users might edit)
    * default log file location
    * data files (non-user files managed completely by the CLI)
* temporarily included code to find old files and copy them; the plan is to leave this in for at least a year after the yargs release
* the `config` command now tells the user where the config file is found
* updated and added unit tests

### DIFF
--- a/src/lib/command/smartthings-command.ts
+++ b/src/lib/command/smartthings-command.ts
@@ -1,14 +1,18 @@
+import envPaths from 'env-paths'
 import log4js from 'log4js'
 import { type Argv } from 'yargs'
 
 import { type CLIConfig, loadConfig, type Profile } from '../cli-config.js'
+import { ensureDir } from '../file-util.js'
 import { buildDefaultLog4jsConfig, loadLog4jsConfig } from '../log-utils.js'
-import { BuildOutputFormatterFlags } from './output-builder.js'
 import { defaultTableGenerator, type TableGenerator } from '../table-generator.js'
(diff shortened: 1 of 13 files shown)


src/lib/command/smartthings-command.ts
```diff
@@ -21,37 +25,101 @@ export const smartThingsCommandBuilder = <T extends object = object>(
 			type: 'string',
 			default: 'default',
 		})
+		.option('verbose-directory-logging', {
+			describe: 'including this will log, to stderr, information about config, data, and' +
+				' logging directories found and used',
```
Sitlintac, line 30: What about this wording?
`log information about config, data, and logging directories to stderr`

Sitlintac (approved):

temporary-notes.md
```diff
@@ -2,3 +2,6 @@ Collected notes to eventually include in the yargs release notes.
 
 * capability version command line arguments have been changed from (in yargs-speak) positionals to
   flags (options in yargs-speak)
+* config file location is now determined by https://www.npmjs.com/package/env-paths[envPaths] library rather.
```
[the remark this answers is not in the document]
rossiam: Oops, rather than oclif. Fixed.

src/lib/command/smartthings-command.ts
rossiam: Heh, that's way better so I updated it. I also added a comment on the option to describe that it is temporary.

PR #678, opened 2025-02-04
refactor: convert installedschema:delete command to yargs

rossiam:
* converted `installedschema:delete` command to yargs
* converted existing tests to yargs
* did not fill out tests for this one; can work on this once all commands are converted


src/commands/installedschema/delete.ts
```diff
@@ -0,0 +1,65 @@
...
+		config.listTableFieldDefinitions.splice(3, 0, 'location')
+	}
+
+	const id = await selectFromList(command, config, {
+		preselectedId: argv.id,
+		listItems: () => installedSchemaInstances(command.client, argv.location, { verbose: argv.verbose }),
+		promptMessage: 'Select an installed schema app to delete.',
+	})
+	await command.client.schema.deleteInstalledApp(id)
+	console.log(`Installed schema app ${id} deleted.`)
```
Sitlintac, line 61: The prompt and log use "schema app" while the help text above uses "schema connector".  This is carryover from the OCLIF version and not a new inconsistency, but also long as we're fiddling with it we should standardize it.

Sitlintac (approved):

src/commands/installedschema/delete.ts
rossiam: I've temporarily included the text `_something_` and marked the PR as draft until we can sort this out.

rossiam: Updated now based on new terms documentation ill use. This includes updating other commands already translated. (See second commit in this PR.)

julieganzer (approved):
LGTM

PR #680, opened 2025-02-06
refactor: convert capabilities:create command to yargs

rossiam:
* converted `capabilities:create` command to yargs
* unit tests to be added later
* there are several TODOs that were carried over from previous code


Sitlintac (approved):
There are some prompts and other user feedback that could be cleaned up, but I don't think it needs to be done in this PR since we'll be coming back to the TODOs eventually.

PR #690, opened 2025-02-24
refactor: convert devices:commands command to yargs

rossiam:
* converted `devices:commands` to yargs
* split some code out into a module
* unit tests will be written later
* (an issue has been written to look into cleaning up some inconsistencies in this command; do unit tests after this)


src/commands/devices/commands.ts
```diff
@@ -0,0 +1,65 @@
...
+		)
+		.example([
+			['$0 devices:commands', 'prompt for a device and details of the command to send'],
+			[
+				'$0 devices:commands 82532e8c-e35b-4fa8-b45f-58a083e5f3d2 switch:off',
+				'send the "off" command for the "switch" capability to the specified device',
+			],
+			[
+				'$0 devices:commands 6d73e4d1-3278-44d4-915a-099a62f05330 \'switchLevel:setLevel(50)\'',
+				'send the "setLevel" command passing 50 as an argument for the "switch" capability to the specified device' +
```
julieganzer, line 39: switchLevel capability

src/lib/command/util/devices-commands.ts
```diff
@@ -0,0 +1,234 @@
...
+			type: 'input',
+			name: 'component',
+			message: 'Enter component index or id',
+			validate: (input: string) => {
+				return inputRegex.test(input) || 'Invalid command syntax'
+			},
+		})).component
+
+		if (isIndexArgument(input)) {
+			cmd.component = device.components[Number.parseInt(input) - 1].id || ''
```
julieganzer, line 92: oh jeez...  and maybe this is totes fine. but is input safe to -1?

rossiam: Probably not but this is just moved code and I've written an issue to address issues with it already.

src/commands/devices/commands.ts
rossiam: Good catch! Fixed.

rossiam: (I also cleaned up the text in the note.)

PR #695, opened 2025-02-26
refactor: convert schema:delete command to yargs

rossiam:
* converted `schema:delete` command to yargs
* made use of `chooseSchemaApp` in command (previously duplicated what `chooseSchemaApp` does)
* updated terminology to use "schema app link"
* converted associated unit tests


Sitlintac (approved):
I really like the "link/unlink" terminology. I think it's clearer what's actually being accomplished.

PR #698, opened 2025-03-05
refactor: convert virtualdevices:create-standard command to yargs

rossiam:
* converted `virtualdevices:create-standard` command and associated libraries to yargs
* updated some unit tests but left others out (this command needs to be updated to use the `input-item` module and tests should wait until after that)
* added several `TODOs` to unconverted code (under `packages`) to be done when converting that code


src/lib/command/util/virtualdevices.ts
rossiam, line 1: I just updated these tests enough to get them working since we should change this to use the `input-item` module.

Sitlintac (approved):

PR #702, opened 2025-03-13
fix: allow 0 for integer values in `askForOptionalInteger`

rossiam:
Fixed a bug where entering the number 0 in a prompt for an optional integer was treating it as if nothing were entered.


packages/lib/src/user-query.ts
```diff
@@ -124,7 +124,7 @@ export const askForOptionalInteger = async (message: string, options?: AskForInt
 		...options,
 		validate: options?.validate ? allowEmptyFn(options.validate) : undefined,
 	}
-	return await promptForInteger(message, updatedOptions) || undefined
```
rossiam, line 127: What the heck was I even thinking when I added this??? 🤦

PR #712, opened 2025-03-27
refactor: convert locations:rooms:create command to yargs

rossiam:
* converted `locations:rooms:create` command to yargs
* updated associated unit tests


src/commands/locations/rooms/create.ts
```diff
@@ -0,0 +1,55 @@
...
+const describe = 'create a room'
+
+const builder = (yargs: Argv): Argv<CommandArgs> =>
+	inputAndOutputItemBuilder(apiCommandBuilder(yargs))
+		.option('location', {
+			alias: 'l',
+			describe: 'the location for the new room',
+			type: 'string',
+		})
+		.example([
```
Sitlintac, line 34: Should we add an example that uses the location option?

Sitlintac (approved):

PR #714, opened 2025-03-31
refactor: convert schema:regenerate command to yargs

rossiam:
* converted `schema:regenerate` to yargs
* updated associated unit tests
* fixed a couple of other minor issues I noticed along the way (in particular, examples for `presentation:device-config` command)


src/commands/presentation/device-config.ts
```diff
@@ -30,14 +30,9 @@ const builder = (yargs: Argv): Argv<CommandArgs> =>
...
-			['$0 apps:settings', 'prompt for an app and get settings for it'],
 			[
-				'$0 apps:settings 3',
-				'get settings for the third app in the list generated by "smartthings apps"',
-			],
-			[
-				'$0 apps:settings 392bcb11-e251-44f3-b58b-17f93015f3aa',
-				'get the settings of the app with the given id',
+				'$0 presentation:device-config 392bcb11-e251-44f3-b58b-17f93015f3aa',
+				'get the device config for the given presentation id',
```
Sitlintac, line 35: Good catch!

src/commands/schema/regenerate.ts
```diff
@@ -0,0 +1,57 @@
...
+			apiDocsURL('generateStOauthCredentials'))
+
+const handler = async (argv: ArgumentsCamelCase<CommandArgs>): Promise<void> => {
+	const command = await apiOrganizationCommand(argv)
+
+	const id = await chooseSchemaApp(
+		command,
+		argv.id,
+		{ promptMessage: 'Select a schema app to regenerate its client id and secret.' },
+	)
```
Sitlintac, line 47: Can/should we repeat the warning about the old values being invalidated (and thus potentially breaking any existing installations) here as well?  Not everyone takes the time to read documentation.

Sitlintac (approved):

src/commands/schema/regenerate.ts
rossiam: That's a good idea but putting it here would only help those who use it in Q&A mode. I think it it would be better prompt for confirmation with the warning. I wrote an issue for it though since that's a bit more than a refactor (even though it is pretty simple).

PR #719, opened 2025-04-07
refactor: convert installedschema command to yargs

rossiam:
* converted `installedschema` command to yargs
* converted relevant unit tests
* fixed a couple of minor errors in tests for `installedapps` command, including removing the old file which has already been rewritten


src/commands/installedschema.ts
```diff
@@ -0,0 +1,90 @@
...
+		tableFieldDefinitions,
+	}
+
+	if (argv.verbose) {
+		config.listTableFieldDefinitions = listTableFieldDefinitions.toSpliced(3, 0, 'locationId', 'location')
+		config.tableFieldDefinitions = tableFieldDefinitions.toSpliced(5, 0, 'location')
+	}
+	const verboseInstalledApp: (app: Promise<InstalledSchemaApp>) => Promise<InstalledSchemaApp & WithNamedLocation> =
+		async app => argv.verbose ? withLocation(command.client, await app) : app
+		// argv.verbose ? async app => withLocation(command.client, await app) : app => app
```
Sitlintac, line 78: I'm having a little trouble following what's happening in lines 76-78.  I think `verboseInstalledApp` is of type `(app: Promise<InstalledSchemaApp>) => Promise<InstalledSchemaApp & WithNamedLocation>`. And whereas the old version would choose between two functions depending on what the value of `argv.verbose` is, now it's technically using the same function but choosing the body depending on the value of `argv.verbose`. So basically a respelling?

Sitlintac (approved):

src/commands/installedschema.ts
rossiam: Sorry, I meant to take out that old one after I made sure the new one worked. I've removed it now.

Yeah, you have the right idea. I would word that second bit slightly differently though. There is now only one function that is making the choice. (I wouldn't say it's "choosing the body" because one function cannot have two bodies. Rather, it has two branches in the same body.)

PR #724, opened 2025-04-17
miscellaneous enhancements to createChooseFn

rossiam:
# miscellanous `createChooseFn` cleanup

* pass through an instance of `APICommand` instead of `SmartThingsClient` which allows functions access to things in `APICommand` (and `SmartThingsCommand`) like the logger
* Added `type` to imports here and there that I noticed were missing it
* fixed a bug where the function created by `createChooseFn` was not using the passed-in `listItems` function
* added support for config defaults to `createChooseFn`
* updated relevant unit tests


src/lib/command/util/util-util.ts
```diff
@@ -29,40 +29,66 @@ export const chooseOptionsDefaults = <T extends object>(): ChooseOptions<T> => (
...
 			preselectedId,
 			autoChoose: opts.autoChoose,
 			listItems: listItemsWrapper,
 			promptMessage: opts.promptMessage,
-		})
+		}
+
+		if (opts.useConfigDefault) {
+			if (!createOptions?.defaultValue) {
+				throw Error('invalid state, choose function called with "useConfigDefault" but no default configured')
```
Sitlintac, line 81: I'm not sure if this change would be appropriate for an error message, but it might be easier to parse if it said something like "choose function was called" because it's a little ambiguous whether "choose" is a verb or an adjective here.  (I.e., that "choose function" is a type of function, not an instruction)

Sitlintac (approved):

PR #734, opened 2025-05-15
refactor: convert edge:drivers:logcat command to yargs

rossiam:
* converted `edge:drivers:logcat` command to yargs
* tried to clean up code some; could still use more work
* updated some unit tests


Sitlintac (approved):
Great work!  Looks good to me.

PR #736, opened 2025-05-19
refactor: convert capabilities:translations:update command to yargs

rossiam:
* convert `capabilities:translations:update` command to yargs


src/commands/capabilities/translations/update.ts
```diff
@@ -0,0 +1,48 @@
...
+	& CapabilityIdInputFlags
+	& InputAndOutputItemFlags
+
+const command = 'capabilities:translations:update [id]'
+
+const describe = 'update a capability translation'
+
+const builder = (yargs: Argv): Argv<CommandArgs> =>
+	inputAndOutputItemBuilder(capabilityIdBuilder(apiOrganizationCommandBuilder(yargs)))
+		.positional('id', { describe: 'device preference id', type: 'string' })
```
Sitlintac, line 23: Capability rather than device preference?

Sitlintac (approved):

src/commands/capabilities/translations/update.ts
rossiam: Oops, fixed by deleting that unnecessary line. The `id` positional is already taken care of by the call to `capabilityIdBuilder` on the preceding line.

PR #738, opened 2025-05-20
refactor: convert locations:modes:(create|delete) commands to yargs

rossiam:
* converted `locations:modes:create` and `locations:modes:delete` commands to yargs


src/commands/locations/modes/delete.ts
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+const command = 'locations:modes:delete [id]'
+
+const describe = 'delete a mode'
+
+const builder = (yargs: Argv): Argv<CommandArgs> =>
+	apiCommandBuilder(yargs)
+		.positional('id', { describe: 'mode id', type: 'string' })
+		.option('location', { alias: 'l', describe: 'a specific location to create the mode in', type: 'string' })
```
Sitlintac, line 21: "to delete the mode from"

src/commands/locations/modes/delete.ts
```diff
@@ -0,0 +1,50 @@
...
+			[
+				'$0 locations:modes:delete --location 5dfd6626-ab1d-42da-bb76-90def3153998',
+				'choose the mode to delete from a list of modes in the specified location',
+			],
+			[
+				'$0 locations:modes:delete 636169e4-8b9f-4438-a941-953b0d617231',
+				'delete the mode with the specified id',
+			],
+		])
+		.epilog(apiDocsURL('createMode'))
```
Sitlintac, line 33: Something along the lines of `deleteMode`?

Sitlintac (approved):

src/commands/locations/modes/delete.ts
rossiam: Oops. Ugh, this is another one of those that's super weird to write correctly. It's not really "to delete the mode from" either though, it's just a filter for when the user doesn't specify a node. I've changed it to the generic "to query" we've been using everywhere else.

src/commands/locations/modes/delete.ts
rossiam: Oops, fixed.

PR #743, opened 2025-05-21
refactor: convert deviceprofiles:translations(|:upsert) commands to yargs

rossiam:
* converted `deviceprofiles:translations` and `deviceprofiles:translations:upsert` commands to yargs


src/commands/deviceprofiles/translations.ts
```diff
@@ -0,0 +1,101 @@
...
+		.option('verbose',
+			{ alias: 'v', describe: 'include list of locales in table output', type: 'boolean', default: false })
+		.example([
+			[
+				'$0 deviceprofiles:translations',
+				'prompt for a device profile and list its locales',
+			],
+			[
+				'$0 deviceprofiles:translations --verbose',
+				'including locales with profiles when prompting for a device profile',
```
Sitlintac, line 50: To be consistent with other example text, I think it would be better for this to say "include locales"

Sitlintac (approved):

src/commands/deviceprofiles/translations.ts
rossiam: Um, I think you mean, "to not sound ridiculous, ...". 😂 Fixed.

PR #746, opened 2025-05-22
refactor: convert invites:schema(|:delete) commands to yargs

rossiam:
* converted `invites:schema` and `invites:schema:delete` commands to yargs


src/commands/invites/schema.ts
```diff
@@ -0,0 +1,93 @@
...
+			[
+				'$0 invites:schema --schema-app=viper_7db10232-3f97-4618-924b-807bf852c616 3',
+				'display details about the third invitation listed in the previous example',
+			],
+			[
+				'$0 invites:schema 97e44afd-845f-4da1-a7b3-fd2625fc9367',
+				'list details of specified invitation',
+			],
+		])
+		.epilog(apiDocsURL('getModes', 'getMode'))
```
Sitlintac, line 55: This seems like the wrong URL for this command.

src/commands/invites/schema/delete.ts
```diff
@@ -0,0 +1,47 @@
...
+		.positional('id', { describe: 'Schema App invitation id', type: 'string' })
+		.option('schema-app', { describe: 'Schema App id', type: 'string' })
+		.example([
+			[
+				'$0 invites:schema:delete',
+				'choose the invitation to delete from a list',
+			],
+			[
+				'$0 invites:schema:delete --schema-app e077532e-3c0f-4f3b-a1d8-029115ee1602',
+				'select an invitation to delete from list, for the specified schema app',
```
Sitlintac, line 30: A bit too long, maybe, but what about "choose the invitation to delete from a list of invitations for the specified schema app"?

src/commands/invites/schema/delete.ts
```diff
@@ -0,0 +1,47 @@
...
+			],
+			[
+				'$0 invites:schema:delete 5dfd6626-ab1d-42da-bb76-90def3153998',
+				'delete the invitation with the specified id',
+			],
+		])
```
Sitlintac, line 36: Should there be a docs URL here?

Sitlintac (approved):

src/commands/invites/schema.ts
rossiam: Oops. Fixed.

src/commands/invites/schema/delete.ts
rossiam: There doesn't seem to be one to add.

src/commands/invites/schema/delete.ts
rossiam: Yeah, that's definitely better and not any longer than some of our other examples so I've used your wording. (And I also capitalized Schema App.)

PR #751, opened 2025-05-28
refactor: convert edge:channels:invites:* commands to yargs

rossiam:
* converted `edge:channels:invites:(create|accept|delete)` commands to yargs


julieganzer (approved):
LGTM

PR #756, opened 2025-05-29
refactor: convert presentation command to yargs

rossiam:
* converted presentation command to yargs

### DIFF
--- a/src/lib/command/util/presentation-table.ts
+++ b/src/lib/command/util/presentation-table.ts
@@ -1,12 +1,20 @@
-import { PresentationDevicePresentation } from '@smartthings/core-sdk'
+import { type PresentationDevicePresentation } from '@smartthings/core-sdk'
 
-import { APICommand, outputItem, TableGenerator } from '@smartthings/cli-lib'
+import { type TableGenerator } from '../../table-generator.js'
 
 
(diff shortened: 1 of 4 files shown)


src/lib/command/util/presentation-table.ts
```diff
@@ -81,46 +89,3 @@ export function buildTableOutput(tableGenerator: TableGenerator, presentation: P
...
-		},
-	]
-
-	static examples = [
-		'$ smartthings presentation fd4adb7f-4a23-4134-9b39-05ed889a03cf',
-		'$ smartthings presentation 4ea31e30-2aba-41c7-a3ec-8f97423d565a SmartThings',
-		'$ smartthings presentation fd4adb7f-4a23-4134-9b39-05ed889a03cf --language=ko',
-		'$ smartthings presentation fd4adb7f-4a23-4134-9b39-05ed889a03cf --language=NONE',
-		'',
-		'Specifying only the presentationId defaults to the "SmartThingsCommunity" manufacturer',
```
dstanich, line 113: Does this info about specifying only the ID and the defaulting of `SmartThingsCommunity` need to be added to the new version?  There is an example that is ID only that mentions `SmartThingsCommunity` but maybe not as explicitly as it does here.

dstanich (approved):

PR #771, opened 2025-06-16
refactor: convert edge:channels:(un)?assign command to yargs

rossiam:
* converted `edge:channels:unassign` and `edge:channels:assign` commands to yargs
* updated relevant unit tests
* updated `chooseChannel` function to support a specified driver id so if the user specified a driver id but no channel, we only present them with channels that have that driver assigned
* updated `chooseChannel` function to use `createChooseFn`


src/commands/edge/channels/assign.ts
```diff
@@ -0,0 +1,64 @@
...
+				'$0 edge:channels:assign --channel 6ea857c6-23d4-4aae-aaf7-7a36daf42f92',
+				'prompt for a driver and assign it to the specified channel',
+			],
+			[
+				'$0 edge:channels:assign 636169e4-8b9f-4438-a941-953b0d617231' +
+					' --channel 6ea857c6-23d4-4aae-aaf7-7a36daf42f92',
+				'assign the specified driver to the specified channel',
+			],
+		])
+		.epilog(apiDocsURL('createDriverChannel'))
```
Sitlintac, line 42: Should there also be an example using `driver-version`?

src/commands/edge/channels/assign.ts
```diff
@@ -0,0 +1,64 @@
...
+		{ promptMessage: 'Select a channel for the driver.', useConfigDefault: true },
+	)
+	const driverId = await chooseDriver(command, argv.driverId, { promptMessage: 'Select a driver to assign.' })
+
+	// If the version wasn't specified, grab it from the driver.
+	const driverVersion = argv.driverVersion ?? (await command.client.drivers.get(driverId)).version
+
+	await command.client.channels.assignDriver(channelId, driverId, driverVersion)
+
+	console.log(`${driverId}${argv.version ? ` (version ${argv.version})` : ''} assigned to channel ${channelId}`)
```
Sitlintac, line 60: Does this want to be `driverVersion` rather than `argv.version`?  Also, if the latter, shouldn't it be `argv.driverVersion`?

src/lib/command/util/edge/channels-choose.ts
```diff
@@ -1,58 +1,52 @@
...
+			const channelsToKeep: string[] = []
+			await Promise.all(allChannels.map(async channel => {
+				const assignedDrivers = await command.client.channels.listAssignedDrivers(channel.channelId)
+				if (assignedDrivers.find(assignedDriver => assignedDriver.driverId === options.withDriverId)) {
+					channelsToKeep.push(channel.channelId)
+				}
+			}))
+			return allChannels.filter(channel => channelsToKeep.includes(channel.channelId))
+		}
+		return allChannels
```
Sitlintac, line 36: Why does it compile a list of `channelId`s and then filter the channel list by that rather than compile a list of `channel`s?

Sitlintac (approved):

src/commands/edge/channels/assign.ts
rossiam: I didn't include one because that is rarely, if ever, needed.

PR #798, opened 2025-11-10
feat: add topic and sub-command info to help output

rossiam:
First of two parts for adding topic and sub-command help to each command's help.

This PR adds:

* adds a `command-util` module to help get topic and sub-command information for a given command.
* adds a `help` module:
  * takes over code generating API docs notes
  * also moved `itemInputHelpText` here which generates help for use with the item-input module
  * added a new `buildEpilog` which will build an epilog for use with yargs' `.epilog` which supports a variety of things we like to put in the help output
* converts a handful of commands to use `buildEpilog`

The second part of this will:

*  convert the remaining commands to use `buildEpilog`
* remove the leftover code in the `api-command` module

### DIFF
--- a/src/__tests__/lib/help.test.ts
+++ b/src/__tests__/lib/help.test.ts
@@ -1,0 +1,149 @@
+import { jest } from '@jest/globals'
+
+import type { BorderConfig, getBorderCharacters, table } from 'table'
+import type { findTopicsAndSubcommands } from '../../lib/command-util.js'
+
+
+const borderConfig = { topLeft: 'top-left' } as BorderConfig
+const getBorderCharactersMock = jest.fn<typeof getBorderCharacters>()
+	.mockReturnValue(borderConfig)
+const tableMock = jest.fn<typeof table>()
+jest.unstable_mockModule('table', () => ({
+	getBorderCharacters: getBorderCharactersMock,
+	table: tableMock,
+}))
+
+const findTopicsAndSubcommandsMock = jest.fn<typeof findTopicsAndSubcommands>()
+	.mockReturnValue({ topics: [], subCommands: [] })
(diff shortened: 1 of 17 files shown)


src/commands/edge/channels/invites/create.ts
```diff
@@ -41,6 +42,7 @@ const builder = (yargs: Argv): Argv<CommandArgs> =>
 				'create an invite from prompted input for the specified channel',
 			],
 		])
+		.epilog(buildEpilog({ command }))
```
rossiam: This is an example of a command that doesn't strictly need an update. Since it's a leaf command and doesn't use any notes or have any API docs, this effectively does nothing. (The `buildEpilog` function in this case returns an empty string which yargs is happy to ignore, or maybe it appends it, either having the same effect. 😄)

I'm kind of leaning towards including this in even these cases but I could be convinced otherwise. The main reason to add it would be that if we don't and a command gets sub-commands added, we will have to remember to update the parent command with an epilog.

src/__tests__/lib/help.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+		// Call this trivial function to fulfill test coverage. :-)
+		expect(tableMock.mock.calls[0][1]?.drawHorizontalLine?.(0, 0)).toBe(false)
+	})
+
+	it('joins sections correctly when multiple present', () => {
+		findTopicsAndSubcommandsMock.mockReturnValueOnce({ topics: ['test::topic'], subCommands: [] })
+
+		expect(buildEpilog({ command: 'test', formattedNotes: 'formatted node' }))
+			.toBe('Notes:\nformatted node\n\nTopics:\n  test::topic')
```
Sitlintac: Note/node here too

PR #806, opened 2025-12-16
chore: update prep for releasing ESM and yargs-based CLI as version 2.0

rossiam:
* ran `npx changeset` to generate changeset info for major update
* removed no-longer-used `tsconfig-test.json` from eslint config
* added a skeletal copilot instruction set (mostly because I was annoyed with it constantly wanting to use single-letter variable names)
* lots of work to ci.yaml and cd.yaml
  * updated action versions
  * removed code handling mono-repo (everything is now just in one npm package)
  * renamed job that publishes npm release to `npm-release`
  * added a bunch of debug; since lots of things changed this will hopefully expedite debugging problems that arise
  * removed installation of QEMU which `pkg` needed but `nexe` does not
  * updated to use auto-generated release notes
* updated `build-binaries` script to put compressed binaries directly in `dist_bin` directory which makes the cd script a little cleaner
* minor updates to changelog text to make it more clear
* updated python version and modules used for functional tests
* fixed functional test that checked `--version` output since the version output is now simpler


.github/workflows/cd.yaml
```diff
@@ -115,111 +111,126 @@ jobs:
...
     runs-on: macos-latest
 
     steps:
     - name: Checkout Repo
-      uses: actions/checkout@v3
+      uses: actions/checkout@v6.0.1
 
     - name: Configure Git Identity
-      uses: Homebrew/actions/git-user-config@master
+      uses: Homebrew/actions/git-user-config@main
```
rossiam: It's gross that we have to use `main` here rather than a specific version. This action doesn't seem to use versions though. We might want to look for alternatives at some point.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Incorrect
+```typescript
+const data = list.map(x => x.id);
+
+try {
+  // ...
+} catch (e) {
+  console.error(e);
+}
```
Sitlintac: Should there be a "Correct" example as well?  Or is that just extra useless tokens for the AI to process?

functional-tests/tests/version_test.py
```diff
@@ -9,4 +9,4 @@ def test_version_output():
         logfile=sys.stdout
     )
 
-    process.expect('@smartthings/cli/.+ .+ node-.+')
+    process.expect('^\\d+\\.\\d+\\.\\d+')
```
Sitlintac: I'm having trouble figuring out what this change is.  Is this because it's not a monorepo anymore?

PR #847, opened 2026-09-01
feat: add hub filter to devices command

### LINKED ISSUE #588: [feat] Add a flag to the `devices` command to filter devices by which Hub they're joined to
**Is your feature request related to a problem? Please describe.**
My specific use-case is probably pretty niche, but I think this could be a valuable filter either way.

I have several locations where I have more than one hub, and I will often have the same devices joined to multiple hubs in the same location. This makes it challenging to determine exactly which device record to isolate when using `smartthings devices`.

Here are two examples of where this happens:

1. I have various LAN devices that will auto-join any time I trigger a "Scan Nearby" flavored discovery. I want to remove them from some of my hubs, but not all of them.
2. I frequently like to test the same device on multiple versions of hub-core by using different hub products. For example, I'll test things on the newest hubcore using my Aeotec v3, and I'll use my TV or ST Station hubs to test slightly older firmware.

**Describe the solution you'd like**
A `-H` flag or something similar to the capital H flag used in other commands to identify a Hub Device UUID that would only return devices that are joined to the given hub.

**Describe alternatives you've considered**

1. Put all of my hubs in different locations and filter by location

This doesn't work great because it's a lot of friction/moving around in the app that slows me down. Also, with more and more products having ST Hub functionality built-in, there is certainly going to be more occurrences for multiple hubs in the same location across the fleet. So there's potential value in getting ahead of that.

2. Doing my own filtering with `jq`

This gets me part way there. For example, I do have a one-liner I sometimes use that looks like this:

`st devices -p production -j | jq -r --arg hub $HUB_UUID '.[] | select(.parentDeviceId == $hub)'`

Where this falls down, though, is that I work with a lot of parent-child devices. This is only a top-level filter, meaning I'd have to craft a gnarly pipeline if I wanted to get all the devices joined to my hub *plus child devices*. Hence, the "entire device tree" in the issue title.

sungmpar:
## Summary

- add a long-form `--hub` filter to `devices` while preserving `-H` for `--health`
- match declared integration-specific hub IDs for DTH, LAN, Matter, Zigbee,
  Z-Wave, Edge Child, and Virtual devices
- include nested child devices by following parent-device ancestry
- preserve existing list filters, numeric-index selection, health, status, and
  verbose output behavior
- add comprehensive regression coverage and a patch changeset

## Root cause

The devices list API has no hub filter, and hub membership can span multiple
parent-device levels. The CLI previously forwarded only API-supported filters
and did not reconstruct device topology, so users with multiple hubs could not
isolate every device joined to one hub, especially nested child devices.

When another list filter narrows the candidates, the implementation loads one
unfiltered device list for ancestry lookup. Missing parents and cycles fail
closed, duplicate candidates are removed, and the selected hub itself is not
returned.

## Verification

- [x] `npm run test -- --runTestsByPath src/__tests__/commands/devices.test.ts --no-watchman` (20 of 20 tests passed)
- [x] `npm run build`
- [x] `node dist/src/run.js devices --help`
- [x] `npm run lint`
- [x] `npm run test -- --no-watchman` (175 suites and 1,577 tests passed)
- [x] `git diff --check`

The built help shows long-form `--hub`, retains `-H` for `--health`, and
includes a hub-filter example. The README generator ran during the build and
produced no tracked README change.

Verification used repository-native mocked devices; no live SmartThings account
or hardware mutation was performed.

Fixes #588

## Checklist

### DIFF
--- a/.changeset/swift-hubs-connect.md
+++ b/.changeset/swift-hubs-connect.md
@@ -1,0 +1,5 @@
+---
+"@smartthings/cli": patch
+---
+
+Add a hub filter to the devices command, including nested child devices.
--- a/src/__tests__/commands/devices.test.ts
+++ b/src/__tests__/commands/devices.test.ts
@@ -5,6 +5,7 @@ import type { ArgumentsCamelCase, Argv, Options } from 'yargs'
 import {
 	type Device,
 	type DeviceHealth,
+	type DeviceListOptions,
 	DeviceHealthState,
 	DeviceIntegrationType,
 	DevicesEndpoint,
@@ -70,6 +71,8 @@ describe('builder', () => {
 	outputItemOrListBuilderMock.mockReturnValue(argvMock)
 
 	const builder = cmd.builder as (yargs: Argv<object>) => Argv<CommandArgs>
+	const optionFor = (name: string): Options | undefined =>
+		(optionMock as OptionMock).mock.calls.find(([key]) => key === name)?.[1]
 
 	it('calls correct parent and yargs functions', () => {
 		expect(builder(yargsMock)).toBe(argvMock)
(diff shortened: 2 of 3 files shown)


sungmpar:
Hi @rossiam, I hope you're doing well.

Whenever you have a chance, would you or someone on the team be able to review this PR? It implements #588, and all CI checks are currently passing.

If everything looks good after review, I'd also appreciate it if it could be merged when convenient. If any changes or additional validation are needed, I'd be happy to make them.

Thank you very much for your time and consideration.

